PR #127, opened 2014-12-23
Update 2013-05-01-rest-lesson-learned-avoid-hackable-urls.html

kijanawoodard:
add comment

### DIFF
--- a/_posts/2013-05-01-rest-lesson-learned-avoid-hackable-urls.html
+++ b/_posts/2013-05-01-rest-lesson-learned-avoid-hackable-urls.html
@@ -192,4 +192,28 @@ tags: [Software Design, Services, REST]
 		</div>
 		<div class="comment-date">2014-10-10 12:02 UTC</div>
 	</div>
+	<div class="comment">
+		<div class="comment-author">
+			<a href="http://kijanawoodard.com/">Kijana Woodard</a>
+		</div>
+		<div class="comment-content">
+			<p>
+				I'd like to propose an alternative strategy to achieve the desired result while maintaining 
+				the ability to easily debug production issues at the client: 
+				only use the non-hackable urls for the "dev sandbox".
+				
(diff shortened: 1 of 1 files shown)


ploeh:
Please accept my apologies that it took me so long to answer, but finally I got around to it: http://blog.ploeh.dk/2013/05/01/rest-lesson-learned-avoid-hackable-urls

kijanawoodard:
No apology necessary.

Fantastic blog. Goldmine.

BTW, I regret my decision to use disqus for comments on my blog and aim to
move toward a model like yours.

ploeh:
FWIW, if you haven't already seen it, this article contains some explanation about my motivation for not choosing Disqus or any other third-party comment engine: http://blog.ploeh.dk/2013/03/03/moving-the-blog-to-jekyll

kijanawoodard:
I went down the "write my own engine" route, but only spent a few hours
based on some code from Tim Thomas. I knew I didn't want a db for the posts
[as you say, let's be real]. I wanted markdown for the posts and easy
hosting.

Easy.

But for comments, I didn't know what to do. I still didn't want my own db,
so I went with disqus [was considering it for a client at the time, but
decided it wasn't a good fit for them - hmmm].

I've had some nice discussions and would rather "own" them rather then
having them "floating in the cloud". I had disqus with my old blog, which I
think was WordPress.  While you can export disqus comments, they don't make
it easy and the comments don't come out cleanly. I had a terrible time
getting from Disqus to Disqus and ended up dropping my original comments.

Needless to say, I regret giving them a second chance.

I'm not sure when I thought of "just make them do pull requests for
comments", but I was kind of laughing about it thinking I was insane. Then
I saw your blog. :-D

Most of my readers will be technical and will likely have github accounts.
I haven't written in a post in quite a while, so haven't taken the time to
modify my "blog engine".

Thanks again for all your writing.

PR #131, opened 2015-01-02
added comment to name operation 'Junction' or 'Consolidation'

feO2x:


ploeh:
Thank you for your comment. Answered on http://blog.ploeh.dk/2014/12/16/name-this-operation

PR #134, opened 2015-01-06
Update 2015-01-06-composition-root-reuse.html

Kennethtruyers:
Added comment


ploeh:
Thank you for writing. Answered at http://blog.ploeh.dk/2015/01/06/composition-root-reuse

PR #135, opened 2015-01-06
added comment regarding reusable composition root code by pushing builde...

feO2x:
...rs or factories to libraries


ploeh:
Thank you writing, and please accept my apologies that it took so long for me to answer; I finally added a response to the page: http://blog.ploeh.dk/2015/01/06/composition-root-reuse

HTH

PR #136, opened 2015-01-06
Added a comment

Skleni:


ploeh:
Thank you writing, and please accept my apologies that it took so long for me to answer; I finally added a response to the page: http://blog.ploeh.dk/2015/01/06/composition-root-reuse

HTH

PR #139, opened 2015-01-15
Correct term for scrolling on wide lines.

davidroberts63:


ploeh:
Nice catch; thank you :)

PR #140, opened 2015-01-15
Commented on "Enumerables are dynamic - also in AutoFixture" 2011-04-18

dcastro:


ploeh:
Thank you for writing; I've answered on

http://blog.ploeh.dk/2011/04/18/EnumerablesaredynamicalsoinAutoFixture

dcastro:
Thank you for the prompt reply!

PR #141, opened 2015-01-16
Add link for functional programming term 'Fold'

Lumirris:
... for those of us still staring wide-eyed at the Functional Programming monolith ;)


ploeh:
Thank you, and sorry about the jargon use :)

PR #142, opened 2015-01-20
Comment - It's also good OO, imho

cessor:
Dear Mark, I wrote a comment under your nice article!


ploeh:
Thank you; I added a short reply on the same page http://blog.ploeh.dk/2015/01/19/from-primitive-obsession-to-domain-modelling

PR #143, opened 2015-01-20
add comment

mcgwiz:


ploeh:
Thank you for your comment; I left a reply on http://blog.ploeh.dk/2015/01/19/from-primitive-obsession-to-domain-modelling

PR #144, opened 2015-01-21
Added Comment to Primitive Object

Suamere:
Added Comment to Primitive Object


ploeh:
Thank you for your comment; I left a reply on the page: http://blog.ploeh.dk/2015/01/19/from-primitive-obsession-to-domain-modelling

PR #145, opened 2015-01-21
added comment to library bundle facade

feO2x:


ploeh:
Thank you for writing; I added a short response.

http://blog.ploeh.dk/2015/01/21/library-bundle-facade

PR #146, opened 2015-01-23
Question: instantiating UserName with lowercase

Lumirris:
Although UserName is stated to be case-insensitive, does this necessarily mean that one must pass all uppercase to the constructor? That doesn't seem right.


ploeh:
Thank you for writing; I added a short response.

http://blog.ploeh.dk/2015/01/19/from-primitive-obsession-to-domain-modelling

PR #147, opened 2015-01-28
Response to a reply to an earlier question

Lumirris:


ploeh:
Thank you for writing; I left a response.

http://blog.ploeh.dk/2015/01/19/from-primitive-obsession-to-domain-modelling

Lumirris:
Thanks, Mark.

I appreciate the reply - lots of good info there for a potential future
blog post.

Cheers,
Jeff


Jeff

PR #148, opened 2015-01-31
Many SCM migrations do not imply the loss of commit messages.

timp21337:


ploeh:
Fair enough :)

FWIW, I've had [AutoFixture](https://github.com/AutoFixture/AutoFixture) go through two migrations so far (TFS->Hg->Git). The first time, we lost all commits before the conversion; the second time, they were essentially preserved.

PR #149, opened 2015-02-01
Added a comment

Skleni:


ploeh:
Thank you for writing; I left a short response

http://blog.ploeh.dk/2015/01/06/composition-root-reuse

PR #150, opened 2015-02-03
Add comment to Explicit Expectations page.

tiesmaster:
The code I added is not formatted, at the moment. I don't know how to do that. Do I need to do that by hand, or do you have some tool, or webapp that can do that?

ps I really enjoy reading your blog, and your book was realy insightful ;)


ploeh:
Thank you for writing; I left a response at http://blog.ploeh.dk/2009/03/11/ExplicitExpectations

Issue #75, opened 2014-06-01
Favicon


ploeh:
The blog should have a favicon...

JoeMighty:
How about one of your face? I'll make one for you :)

ploeh:
Heh, isn't that a little too self-centred?

JoeMighty:
I guess some might see it that way. Though I would argue that some people (including myself in this instance) come to your website because they value your opinion. So therefore you're the brand that your website is promoting. And Scott Hanselman has his face as an favicon! ;)

If not your face then a small P for Ploeh?

ploeh:
If you take the letters from *ploeh* and squash them all together, and then subsequently ignore the *h*, you get something like this:

![ploeh glyph](https://cloud.githubusercontent.com/assets/242165/5852447/27da08f6-a217-11e4-87c5-c9e423abfc42.jpg)

If you then tilt it 45°, it fits pretty well within a square:

![ploeh glyph tilted](https://cloud.githubusercontent.com/assets/242165/5852452/47fb18dc-a217-11e4-8633-9ac6887a4de8.jpg)

That's just about as obscure as the word *ploeh* in itself. Would that work?

feO2x:
What is the meaing of "Ploeh"? It is not a danish word as far as I can tell. If it has a special meaning, this could be incorporated into your favicon.

ploeh:
I could tell you, but then I'd have to kill you ;)

feO2x:
Ok, let's leave it at that :-)

feO2x:
Maybe something like this?

![ploeh favicon](https://cloud.githubusercontent.com/assets/7884158/5853370/56ff675a-a223-11e4-9c92-85e2bbaf4a69.png)

ploeh:
Considering that the most common size users are going to see is a 16x16 pixel image, I'm not too concerned about how the eye scans the icon.

My thought with the above was to have the line go from corner to corner; otherwise, I'm afraid it'd look weird. However, I do grant that it may look strange with half of the icon being empty, so perhaps it's better just to scrap that entire concept...

feO2x:
I wouldn't necessarily say so:
![ploeh favicon v0 2](https://cloud.githubusercontent.com/assets/7884158/5855316/34465a3e-a237-11e4-8a29-3f0c22f537fd.png)

This is how it looks in 16x16:
[...]

PR #151, opened 2015-02-08
added favicons and corresponding links in default.html pages

feO2x:
As discussed in https://github.com/ploeh/ploeh.github.com/issues/75#issuecomment-73397926 I added the favicon in a folder called favicons. I also added links to the favicons in the default.html pages in _includes/themes subfolders.

Unfortunately, I have no experience with Jekyll Bootstrap and had no time to set it up on my own computer, thus the last commit is not tested at all. 

Here are thinks we should consider:
- I didn't use BASE_PATH when setting the hrefs for the links (but they did not do this in their example code either). Is that correct?
- IE10 is currently not supported, i.e. it will not show the favicon (according to this [Stack Overflow Answer] (http://stackoverflow.com/questions/23849377/html-5-favicon-support/23851464#23851464) this isn't really a problem, because most people use IE 11+).


ploeh:
Thank you! It's now live :smile:

PR #153, opened 2015-02-17
Fixed the "web jobs" to webjobs - added survey request

jaime-espinosa:


ploeh:
Thank you for your contribution.

In its current state, I can't, or will not, accept it, since it has two issues:

- Due to the way Jekyll works, one can't just change the title of a post, since the post URL is derived from the title, and that would change the permalink for the post. While there are ways around this, I don't think the proposed change warrants that, since it involves more moving parts.
- The survey link is added as though it's a part of the article; it looks like something that I wrote, so it looks like the survey is my recommendation. Also, a permanent web fixture such as a blog article isn't, I think, a proper place to post links to web resources of more ephemeral nature, and assume that the survey is going to run for some time, and then end.

jaime-espinosa:
Ok, fair enough.  Please add the survey in a way that you feel is most appropriate.  If we are to push this feature forward, we need the support of its users.

ploeh:
This particular article is already old, so it doesn't attract that many viewers; case in point: it had 32 page views the last month, so if you want people to take the survey, you're not going to get much out of adding the link here.

Twitter would be a much better medium for this. If you post a good tweet with a link to the survey, I can retweet it.

jaime-espinosa:
Done.

Jaime Espinosa ‏@jaime3espinosa <https://twitter.com/jaime3espinosa>  6h6 hours ago<https://twitter.com/jaime3espinosa/status/567503110889684992>
Please fill out the survey on Azure WebJobs. We really need the feedback. https://microsoft.qualtrics.com/SE/?SID=SV_8iRbWLZVMIhjqTj …<https://t.co/TaOgaOXn0D> Thank you!

PR #157, opened 2015-03-17
13 is actually 13 not fizz.

ghuntley:
Found a small mistake on your inline test data.


ploeh:
It seems the original kata description is down, but here are the details pulled from the internet archive: https://web.archive.org/web/20130604001354/http://codingdojo.org/cgi-bin/wiki.pl?KataFizzBuzz

The post in question implements stage 2 of the kata, where 13 *should* Fizz.

ploeh:
BTW, sorry if the above comment was a bit curt; I was on my way out the door when I wrote it, although that's no excuse.

While I had to decline this particular pull request, I do appreciate your interest and effort.

ghuntley:
No offense taken :dancer:

PR #158, opened 2015-03-18
Add comment on Domain modelling blog post

mrvux:


ploeh:
Thank you for your comment. It seems that we essentially agree, so I didn't add an explicit answer per se :smile:

PR #159, opened 2015-04-02
Add comment to the avoid 204 responses post

theshadow:
Comment disagreeing with your statement about DELETE and 204.


ploeh:
Thank you for writing. I left a comment on the page http://blog.ploeh.dk/2013/04/30/rest-lesson-learned-avoid-204-responses

PR #160, opened 2015-04-10
Update 2015-01-19-from-primitive-obsession-to-domain-modelling.html

westonized:
Comment on the Equality method used.


ploeh:
Thank you for writing! I've left a short response on the page: http://blog.ploeh.dk/2015/01/19/from-primitive-obsession-to-domain-modelling

PR #161, opened 2015-04-11
Update 2013-05-01-rest-lesson-learned-avoid-hackable-urls.html

dkubb:
Add a comment about recent changes I made to make URLs unhackable in a REST api.


ploeh:
Thank you for writing; I left a response on the page: http://blog.ploeh.dk/2013/05/01/rest-lesson-learned-avoid-hackable-urls

PR #162, opened 2015-04-11
Combatting client side url construction

dkubb:
This branch adds a comment on how we planned to combat client side URL construction and signing.


ploeh:
Thank you for writing; I left another response on the page: http://blog.ploeh.dk/2013/05/01/rest-lesson-learned-avoid-hackable-urls

PR #164, opened 2015-04-13
Fixed Typos

bartelink:
I know const isnt water tight but idiomatic C++ does use const to 'guarantee no changes'


ploeh:
Thank you.

PR #165, opened 2015-04-14
added a comment

chrisdew:


ploeh:
Thank you for writing - you were close enough with the format, but I've added your name and a link. If you're unhappy with either, you're welcome to send another PR :smile: 

Will respond in a while.

ploeh:
Left a short response on the page http://blog.ploeh.dk/2015/04/13/less-is-more-language-features

PR #167, opened 2015-04-18
Update 2015-04-15-c-will-eventually-get-all-f-features-right.html

vkhorikov:


ploeh:
Thank you for writing. I left a response on the page: http://blog.ploeh.dk/2015/04/15/c-will-eventually-get-all-f-features-right

PR #168, opened 2015-04-22
Fixed cited C# 5 release date

bartelink:
As I'm confident you're aware, Async/await came in C# 5 which was in VS 2012 which was in September 2012


ploeh:
Thank you! I honestly couldn't remember when it was :flushed:

PR #170, opened 2015-05-15
Another comment on "less is more"

shirok:


ploeh:
Thank you for writing; I left a response on the page: http://blog.ploeh.dk/2015/04/13/less-is-more-language-features

Issue #77, opened 2014-06-01
Smaller RSS feed


ploeh:
[The RSS feed](http://blog.ploeh.dk/rss.xml) is getting longer and longer, and is currently at over 4 MB. This is not long-term sustainable, as the size will just keep growing.

Unlike Atom, I'm not sure if RSS enables pagination. If it does, the RSS feed should be [paged like the Atom feed](https://github.com/ploeh/ploeh.github.com/issues/76). If it doesn't, I think the most common approach is to truncate the feed so that it only lists the *n* latest posts.

ploeh:
Resolved by #172

Issue #76, opened 2014-06-01
Paged Atom feed


ploeh:
[The Atom feed](http://blog.ploeh.dk/atom.xml) is becoming longer and longer, and is currently at more than 4 MB. This is not long-term sustainable.

The Atom specification enables paged feeds, and Jekyll also enables pagination, so it would be nice to split the Atom feed into multiple, smaller pages.

One challenge, though, is that ideally, once a page is filled up, it should retain its unique ID, but I'm not sure if that's a practical concern...

PR #172, opened 2015-05-18
Limiting Atom and RSS feeds to 10 posts

xdaDaveShaw:
As discussed on twitter and mentioned via my [blog](http://taeguk.co.uk/blog/dealing-with-newsblur-and-large-feeds/), NewsBlur has problems with large RSS / Atom feeds. Jekyll supports limiting the number of posts in the feed. I've picked 10 as an arbitrary number.

### DIFF
--- a/_includes/themes/ploeh/atom.xml
+++ b/_includes/themes/ploeh/atom.xml
@@ -11,7 +11,7 @@
  <subtitle>{{ site.tagline }}</subtitle> 
  <id>{{ site.production_url }}</id>
 
- {% for post in site.posts %}
+ {% for post in site.posts limit:10 %}
  <entry>
  	<title>{{ post.title }}</title>
    	<link rel="alternate" type="text/html" href="{{ site.production_url }}{{ post.url }}"/>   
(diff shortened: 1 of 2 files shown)


ploeh:
Thank you for your contribution. As explained in #76, Atom feeds should be paged. As described in #77, it would also be nice if the RSS feed is, but I'm not sure if that's doable.

Assuming that it isn't possible paging RSS feeds, we can move forward with the RSS change, but please either revert the change to the Atom feed (by far the easiest solution), or implement paging.

xdaDaveShaw:
Doh! I never thought to look into if you have anything logged for this - I've only just discovered Jekyll and found this setting via Nick Craver's blog. I've reverted the Atom changes, no idea how I would do that with Jekyll or Liquid. Looking at the [RSS Spec](https://validator.w3.org/feed/docs/rss2.html) there's nothing supporting paging.

ploeh:
Thank you! It's now live on the blog. Hope it helps :smile:

xdaDaveShaw:
NewsBlur seems to have picked it up. I'll run my Yahoo Pipes along side the RSS until your next blog post to see if it works.

Thanks

PR #173, opened 2015-06-12
Update 2013-04-30-rest-lesson-learned-avoid-204-responses.html

tuespetre:
Added comments


ploeh:
Thank you for writing. I left a response on the page: http://blog.ploeh.dk/2013/04/30/rest-lesson-learned-avoid-204-responses

(Sorry that it took me so long to merge; usually, I do it faster.)

PR #174, opened 2015-06-30
Update 2015-06-22-rest-implies-content-negotiation.html

grahamrhay:


ploeh:
@grahamrhay, please accept my apologies that it took me so long to answer, but I've now left a response on the page: http://blog.ploeh.dk/2015/06/22/rest-implies-content-negotiation

PR #175, opened 2015-07-01
comment on 5 levels of media type

nikpetroff:
Thanks for blogging Mark. Your abstract interfaces blew my mind!
In my comment there is something useful about content negotiation you may not have seen.


ploeh:
Sorry that it took me so long to respond. I've now read the linked article, but must admit that I don't understand what problem it attempts to solve :flushed:

PR #176, opened 2015-07-16
Add comment to 2015-06-22-rest-implies-content-negotiation.html

evmcl:


ploeh:
Thank you for writing; I've left a response on the page: http://blog.ploeh.dk/2015/06/22/rest-implies-content-negotiation

PR #177, opened 2015-07-23
LINQ in GetCalculator

JohnCastleman:
FYI, `IEnumerable<T>.OfType<T>()` method:

    from c in candidates.OfType<IHandleShippingMethod>()
    where c.CanHandle(shippingMethod)
    select c

Also, the following approach might allow multiple selection strategies, depending on use case:

    public IEnumerable<IBasketCalculator> GetAllCalculators(ShippingMethod shippingMethod)
    {
        return from c in candidates.OfType<IHandleShippingMethod>()
               where c.CanHandle(shippingMethod)
               select c;
    }

    public IBasketCalculator GetBestCalculator(ShippingMethod shippingMethod)
    {
        return GetAllCalculators(shippingMethod)
               .First();
    }

Leading to the following possibilities (you have to imagine something other than cost calculations for some of these):

    // use case: the original application
    var cost = factory.GetBestCalculator(shippingMethod).CalculatePrice();

    // use case: average of all results
    var average = factory.GetAllCalculators(shippingMethod).Select(c => c.CalculatePrice()).Average();

    // use case: redundant calculators - all should return same result (ideally, each enum value would have two or more matching calculators)
    var redundancyCheck = factory.GetAllCalculators(shippingMethod).Select(c => c.CalculatePrice()).Distinct().Count() == 1;


ploeh:
Did you mean to add this as a comment to the blog?

JohnCastleman:
Yes, I just threw into into the commit comment and used markdown's formatting.

PR #181, opened 2015-08-06
Versioning based on branch names

gusvmx:


ploeh:
Thank you for writing; I left a response on the page: http://blog.ploeh.dk/2013/12/10/semantic-versioning-with-continuous-deployment

PR #185, opened 2015-09-07
Fix minor typos

bartelink:


_posts/2015-08-10-type-driven-development.html
```diff
@@ -44,7 +44,7 @@
     My problem was that I didn't know how long it took to handle a message, so the Polling Consumer would have to measure that as well, constantly updating its estimates of how long it takes to handle a message.
   </p>
   <p>
-    This wasn't a harder problem that I could <a href="http://blog.ploeh.dk/2014/09/25/faking-a-continuously-polling-consumer-with-scheduled-tasks">originally solve it in a highly coupled imperative fashion</a>. I wasn't happy with that implementation, though, so I'm happy that there's a much more <em>incremental</em> approach to the problem.
+    This wasn't a sufficiently hard problem to thwart my <a href="http://blog.ploeh.dk/2014/09/25/faking-a-continuously-polling-consumer-with-scheduled-tasks">originally solving it in a highly coupled imperative fashion</a>. I wasn't happy with that implementation, though, so I'm happy that there's a much more <em>incremental</em> approach to the problem.
```
ploeh: Doesn't the change to a gerund transform the sentence from active voice into a passive(r) voice?

_posts/2015-08-10-type-driven-development.html
```diff
@@ -74,7 +74,7 @@
     When in the <em>Ready</em> or <em>No message</em> states, the Polling Consumer may also realise that time is running out, and so decide to quit, in which case the (final) state is the <em>End</em> state.
   </p>
   <p>
-    This means that the Polling Consumer needs to measure what it does, and those measurements influence what it decides to do.
+    This means that the Polling Consumer needs to measure what it does, and have those measurements influence what it decides to do.
```
ploeh: I could also have written the original sentence:

*This means that the Polling Consumer needs to measure what it does; those measurements influence what it decides to do.*

I think that introducing "have" changes the meaning of the sentence ever so slightly. Not that it matters much. Do you think your proposed change makes it *easier* to understand the sentence?

_posts/2015-08-10-type-driven-development.html
bartelink: It may well do. To be honest, I'm simply unable to divine what it is that you were trying to say in the first instance (I retained *hard* in an attempt to retain the intent but suspect changing it to something else (complex?) might serve to further mangle it) and hence I very definitely going out on a limb putting those words in your mouth :)

I have no concrete solution :(

_posts/2015-08-10-type-driven-development.html
bartelink: Here I was attempting to bring out the fact that you felt that there was a need to time things, and that those timings would serve as inputs to that influencing the decision. When I read the initial form, that linkage is not there; it's as if the act implied "measure" directly "influence"s the decision.

Your proposal above retains the same weakness in my view hence I wouldn't see it as a meaningful improvement.

_posts/2015-08-10-type-driven-development.html
jamesjrg: FWIW "I could originally solve" is kind of a strange combination of words regardless of the rest of the sentence. That exact phrase has no google hits other than this blog post, and even "could originally solve" is only used in sentences with the sense of somethings abilities changing over time rather than a choosing to do something differently.

Not quite sure why I'm writing a long paragraph on exact choices of words in a blog post, I just happened to click the "open pull requests" button.

FWIW I'd suggest "This is a sufficiently easy problem that I originally used a highly coupled, imperative solution", just because it's shorter and avoids the maybe-double-negative "wasn't sufficiently hard to prevent", but then again colorless green ideas sleep furiously.

ploeh:
Closing due to inactivity.

PR #188, opened 2015-09-21
fixed typo

wolfascu:
Hi Mark!

I spotted a small typo in your latest blog post and here's the fix.

Thanks for the post! The Estimator is indeed a good example.


ploeh:
Thank you :+1:

PR #194, opened 2015-09-29
CQS reference warning

rManiks:
Comment providing reference to CQ(R)S usage warning.


ploeh:
Thank you for your comment. I'll be happy to publish it, but before I do, I thought I'd better ask.

It looks to me as though you're mixing up CQS and CQRS. The article explicitly talks about CQS, and never about CQRS.

If you still want me to publish the comment, please let me know.

rManiks:
Hi Mark,

Yes, they are different and address different problems. (CQS - addresses logical correctness in terms of side effects), while CQRS addresses resource contention (objects, database level) to maximise availability.

I just thought of adding the word of caution in there, as CQS can lead to CQRS. Perhaps I could've been clearer.

Ta,
Mani.

PR #195, opened 2015-10-03
added comment about internal violating the OCP

feO2x:


ploeh:
Thank you for writing; I've left a response on the page: http://blog.ploeh.dk/2015/09/21/public-types-hidden-in-plain-sight

PR #196, opened 2015-10-03
added question about when internal is beneficial

feO2x:


ploeh:
Thank you for writing; I've attempted to answer on the page: http://blog.ploeh.dk/2015/09/22/unit-testing-internals

PR #198, opened 2015-10-07
Add comment on service locator post

dbohdan:


ploeh:
Thank you for writing; I left a response on the page: http://blog.ploeh.dk/2010/02/03/ServiceLocatorisanAnti-Pattern

PR #200, opened 2015-10-09
Comment added by Pvlerick

Pvlerick:


ploeh:
Thank you for writing; I left a response on the page: http://blog.ploeh.dk/2015/10/08/command-query-separation-when-queries-should-have-side-effects

PR #201, opened 2015-10-09
Commented CQS when Queries should have side-effects

gnugat:
Another option would be to send a notification, but it is at the cost of some indirection.


ploeh:
Thank you for writing; I left a response on the page: http://blog.ploeh.dk/2015/10/08/command-query-separation-when-queries-should-have-side-effects

PR #202, opened 2015-10-09
adding a comment

iamkoch:


ploeh:
Thank you for writing, and sorry it took so long to respond; I'm currently travelling with limited access to internet.

PR #206, opened 2015-11-06
Added comment

tonyj444:


ploeh:
Thank you for writing; I've now added a response to the page: http://blog.ploeh.dk/2013/09/11/di-and-events-composition

HTH

PR #208, opened 2015-11-16
Typo: Pragma disable/restore mismatch

PMBjornerud:
This looks like a typo, assume it should be 618 both places instead of 681 on the restore.


ploeh:
Thank you :+1:

PR #211, opened 2015-11-19
Add comment to "Service Locator violates encapsulation"

japanjeff:
Added comment asking if usage of Service Locator is only an anti-pattern if used in certain contexts. First comment on your blog... let me know if there's anything lacking.


ploeh:
Thank you for writing; I've left a response on the page: http://blog.ploeh.dk/2015/10/26/service-locator-violates-encapsulation

HTH

japanjeff:
Thanks for the response.

PR #212, opened 2015-11-21
Adding a comment

rahulpnath:
Adding a comment to http://blog.ploeh.dk/2015/01/19/from-primitive-obsession-to-domain-modelling/


ploeh:
Thank you for writing; I left a response on the page http://blog.ploeh.dk/2015/01/19/from-primitive-obsession-to-domain-modelling

HTH

rahulpnath:
Thank you!

Rahul
http://rahulpnath.com

PR #213, opened 2015-12-03
Adding Comment

riplikash:
Adding a comment. Have no idea if this is how you intended it to be done. :)


ploeh:
Thank you for writing; I've left a response on the page: http://blog.ploeh.dk/2014/01/29/nuget-package-restore-considered-harmful

PR #220, opened 2016-01-06
added comment to "Service Locator violates SOLID"

feO2x:


ploeh:
Thank you for writing; I've left a response on the page: http://blog.ploeh.dk/2014/05/15/service-locator-violates-solid

HTH

PR #221, opened 2016-01-08
added another comment to "Service Locator violates SOLID"

feO2x:
I added another comment explaining why clients should be aware when they target an O-O or Metaprogramming API.


ploeh:
Thank you for writing; I left a response on the page: http://blog.ploeh.dk/2014/05/15/service-locator-violates-solid

PR #222, opened 2016-01-21
comment added

jonathanayoub:


ploeh:
Thank you for writing; I left a response on the page: http://blog.ploeh.dk/2014/06/13/passive-attributes

HTH

PR #224, opened 2016-02-10
Update 2014-01-29-nuget-package-restore-considered-harmful.html

kimsagro1:


ploeh:
Thank you for your comment, although it's hardly good new :disappointed:

PR #227, opened 2016-02-25
Add comment about hackable URIs.

RubenVerborgh:
Thanks for considering my comment, even though it's quite late.

### DIFF
--- a/_posts/2013-05-01-rest-lesson-learned-avoid-hackable-urls.html
+++ b/_posts/2013-05-01-rest-lesson-learned-avoid-hackable-urls.html
@@ -327,4 +327,13 @@ tags: [Software Design, Services, REST]
 		</div>
 		<div class="comment-date">2015-04-12 12:16 UTC</div>
 	</div>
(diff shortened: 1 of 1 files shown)


ploeh:
Thank you for writing, and please accept my apologies that it took me so long to respond. I've left a response on the page: http://blog.ploeh.dk/2013/05/01/rest-lesson-learned-avoid-hackable-urls

RubenVerborgh:
Great :-)

Seems you still have a semi-hackable URI then with `http://foo.ploeh.dk/0D89C19343DA4ED985A492DA1A8CDC53/{term}`. Even though the `/0D89C19343DA4ED985A492DA1A8CDC53/` part is non-manipulable, terms can still be changed at the end. This shows that hackable URIs cannot entirely be avoided if we want to keep hypermedia controls simple. What else are hackable URIs but underspecified, out-of-band hypermedia controls?

RubenVerborgh:
<= should probably make this into a pull request, let me do that

PR #229, opened 2016-03-02
Reply to comment on URI templates

RubenVerborgh:
Following up on #227.


RubenVerborgh:
…now I'm very curious to the reply :wink:

ploeh:
I've posted a reply on the page: http://blog.ploeh.dk/2013/05/01/rest-lesson-learned-avoid-hackable-urls

HTH

PR #230, opened 2016-03-10
Comment on using POST to avoid hackability

RubenVerborgh:
I argue that hackable URLs should not be avoided at any cost, especially not if the workaround makes the API more complex.


ploeh:
Thank you for writing. I've left a response on the page: http://blog.ploeh.dk/2013/05/01/rest-lesson-learned-avoid-hackable-urls

RubenVerborgh:
Thanks for the reply, that nicely ends the discussion :+1:

PR #231, opened 2016-03-18
Added comment to post functional architecture

baks:
Added comment to post about functional architecture


ploeh:
Thank you for writing. I left a response on the page: http://blog.ploeh.dk/2016/03/18/functional-architecture-is-ports-and-adapters

PR #234, opened 2016-04-06
Followup Comment

VTJDailey:
Hi Mark,

Here's a brief followup comment if you're interested


ploeh:
Thank you for writing. I'm currently travelling, so it may take some time before I have time to respond, but I'll get around to it eventually.

ploeh:
I've left a response on the page: http://blog.ploeh.dk/2016/03/18/functional-architecture-is-ports-and-adapters

PR #236, opened 2016-04-13
[fix] link to 'about' page

jpcaruana:
the page http://blog.ploeh.dk/about.html doesn't exist, but http://blog.ploeh.dk/about/ does exist


ploeh:
Thank you :+1:

jpcaruana:
thanks to **you**: I took your /tags page for my own blog.

PR #237, opened 2016-04-16
Commented on "Async as surrogate IO"

dcastro:


ploeh:
Thank you for writing; I left a response on the page: http://blog.ploeh.dk/2016/04/11/async-as-surrogate-io

dcastro:
Thank you very much for taking the time to reply! I left a response here: #238

PR #239, opened 2016-05-06
Comment to CQS and server-generated Entity IDs

vkhorikov:
Hopefully, managed the mark-up correctly.


ploeh:
Thank you for writing. I left a response on the page: http://blog.ploeh.dk/2016/05/06/cqs-and-server-generated-entity-ids

PR #241, opened 2016-05-10
comment on cqs

cocogorilla:


ploeh:
Thank you for writing; I added a response on the page: http://blog.ploeh.dk/2016/05/06/cqs-and-server-generated-entity-ids

PR #243, opened 2016-05-25
added comment

eugene-g:
Added my comment and fixed typo in the timestamp of previous comment.


ploeh:
Thank you for writing; I left a response on the page: http://blog.ploeh.dk/2015/01/19/from-primitive-obsession-to-domain-modelling

PR #245, opened 2016-06-09
comment about using Action instead of IObserver.

SpencerJB:
Comment on DI and Events series of posts.


ploeh:
Thank you for writing. I left a response on the page http://blog.ploeh.dk/2013/09/11/di-and-events-composition

SpencerJB:
Thank you for taking the time to reply to my query.

PR #247, opened 2016-06-23
Adding a comment to 2013-05-01-rest-lesson-learned-avoid-hackable-url…

AshleyAitken:
…s.html

Interesting blog implementation.  It must keep away a lot of commenters. I don't know what I am doing but I enjoyed the challenge  :-)


ploeh:
Thank you for writing; I've left a response on the page: http://blog.ploeh.dk/2013/05/01/rest-lesson-learned-avoid-hackable-urls

HTH

PR #249, opened 2016-07-04
Added a comment about effect typing

benjamin-hodgson:


ploeh:
Thank you for writing. There's lot of content there for me to contemplate (I never claimed to be a Haskell expert). Apart from that, I don't think that I currently have anything to say in response, other than thank you for pointing me to more things I have to learn :smile:

PR #250, opened 2016-07-04
Add my comment to conditional-composition post

MartinRykfors:
Hi Mark,

Many thanks for this post. I've added my comments at the end, and I think I've learned a lot from this exchange!


ploeh:
Thank you for writing; I left a short reply on the page: http://blog.ploeh.dk/2016/07/04/conditional-composition-of-functions

PR #253, opened 2016-09-11
Added a comment to an older blog post

ardave:


ploeh:
Thank you for writing :+1: I left a short response on the page http://blog.ploeh.dk/2011/04/29/Feedbackmechanismsandtradeoffs

PR #255, opened 2016-09-27
Added Result union type,

moodmosaic:
so that the `Result` module compiles.

Hopefully, this makes it easier for readers to just copy and paste, and compile the source code using their favorite code editor or IDE.


ploeh:
Thank you! I must have forgotten to select the `Result` type itself when I copied the code over 😳

PR #256, opened 2016-10-14
Master

blainne:
A question to "Decoupling decissions from effects".


ploeh:
Thank you for writing. I've left a response on the page http://blog.ploeh.dk/2016/09/26/decoupling-decisions-from-effects

PR #257, opened 2016-11-16
Added comment to 2014-06-13-passive-attributes

TyreeJackson:


ploeh:

PR #258, opened 2016-12-03
Add comment on Easy domain modeling with types

balazsbotond:
I added a comment on your latest article.


ploeh:
Thank you for writing; I've left a response on the page: http://blog.ploeh.dk/2016/11/28/easy-domain-modelling-with-types

HTH

PR #261, opened 2017-01-20
Commenting 2016-03-18-functional-architecture-is-ports-and-adapters.html

giuliohome:
This is a follow up of my twitter's comments about I/O in pure functions and the concrete benefits of Haskell-like purity


ploeh:
Thank you for writing. I left a brief response on the page: http://blog.ploeh.dk/2016/03/18/functional-architecture-is-ports-and-adapters

PR #262, opened 2017-01-20
comment to 2016-03-18-functional-architecture-is-ports-and-adapters.html

giuliohome:
I've posted a reply to better clarify my initial comments


ploeh:
Thank you for writing. I left a response on the page http://blog.ploeh.dk/2016/03/18/functional-architecture-is-ports-and-adapters

PR #265, opened 2017-02-02
Benefit of `tryAcceptComposition`?

dustinmoris:
Hi, Thank you for those blog posts. I also read your other posts on ports and adapters and the proposed architecture makes sense in terms of how it works, but I struggle to see the benefit in a real world application. Let me maybe explain with a quick example.

In the 2nd blog post you demoed this function:

```
// int -> (DateTimeOffset -> Reservation list) -> (Reservation -> int) -> Reservation
// -> int option
let tryAccept capacity readReservations createReservation reservation =
    let reservedSeats =
        readReservations reservation.Date |> List.sumBy (fun x -> x.Quantity)
    if reservedSeats + reservation.Quantity <= capacity
    then createReservation { reservation with IsAccepted = true } |> Some
    else None
```

If I understand it correctly this function is pure if `readReservations` and `createReservation` are both pure as well and it is impure otherwise.

The benefit of having a pure function is because it is extremely easy to understand, test and reason about, that makes sense.

So then in the 3rd blog post you showed how to make `tryAccept` pure, by removing the function dependencies and replacing it with values:

```
// int -> Reservation list -> Reservation -> Reservation option
let tryAccept capacity reservations reservation =
    let reservedSeats = reservations |> List.sumBy (fun x -> x.Quantity)
    if reservedSeats + reservation.Quantity <= capacity
    then { reservation with IsAccepted = true } |> Some
    else None
```

However, additionally you had to move the impure code into a new function:

```
// Reservation -> int option
let tryAcceptComposition reservation =
    reservation.Date
    |> DB.readReservations connectionString
    |> flip (tryAccept 10) reservation
    |> Option.map (DB.createReservation connectionString)
```

So after all the application hasn't reduced the total number of impure functions (still 3 in each case - readReservations, createReservation and tryAccept[Composition]). The only difference is that one impure function `tryAccept` has been refactored into one pure function and a second impure function. What is the benefit of that? If I still remain with the same amount of impure functions and all of them are 100% testable from a unit testing point of view, did I win anything by making that extra split? I find the original `tryAccept` much easier to read and understand than the combination of `tryAccept` + `tryAcceptComposition`.

Am I missing something?


ploeh:
Thank you for a great question. Would you mind adding this as a question to the article itself, so that other people could benefit from the answer?

If you don't care to deal with [my idiosyncratic commenting system](https://github.com/ploeh/ploeh.github.com#comments), I can do it for you by copying the text of your comment. I wouldn't do that without your permission, though.

dustinmoris:
I have moved it, but writing a question in HTML markup was not fun :(

ploeh:
> writing a question in HTML markup was not fun

I know, sorry about that.

On the other hand, [your question is now live on the page](http://blog.ploeh.dk/2017/02/02/dependency-rejection), and it's looking great!

It's Friday evening for me know, so it may be a few days before I get around to writing an answer, but I'll post a notice here when I've done so.

ploeh:
I've added a response to the page: http://blog.ploeh.dk/2017/02/02/dependency-rejection

HTH

PR #266, opened 2017-02-04
Add comment about Json serialization of records

markvincze:
Comment with a question about how the Json serialization of immutable F# records work.


ploeh:
Thank you for writing. I've left a response on the page http://blog.ploeh.dk/2013/10/15/easy-aspnet-web-api-dtos-with-f-climutable-records

HTH

PR #268, opened 2017-02-06
Using SQL Server sequence

ndc:
We can use SQL Server sequence as running number generator service.


ploeh:
Thank you for your comment 👍

PR #269, opened 2017-02-06
Added a comment to the Dependency Rejection post

benjamin-hodgson:


ploeh:
Thank you for writing, and please accept my apologies for the late reply. I have, however, finally left one on the page: http://blog.ploeh.dk/2017/02/02/dependency-rejection

PR #270, opened 2017-02-07
Question about HttpStatusCodes in domain model.

wortho:
The PR adds a comments section and asks a question in the comments.


ploeh:
Thanks for writing; I've left a response on the page: http://blog.ploeh.dk/2017/01/03/decoupling-application-errors-from-domain-models

HTH

PR #271, opened 2017-02-28
commented 2014-06-02-captive-dependency.html

bitbonk:
Added a comment about how to solve the captive dependency proplem.


ploeh:
Thank you for your comment.

PR #272, opened 2017-03-12
comment on functional-architecture-is-ports-and-adapters

beluchin:


ploeh:
Thank you for writing. I've left a response on the page: http://blog.ploeh.dk/2016/03/18/functional-architecture-is-ports-and-adapters

HTH

PR #273, opened 2017-03-19
Added comment

techyourchance:


ploeh:
Thank you for writing. I've left a response on the page: http://blog.ploeh.dk/2011/03/04/Composeobjectgraphswithconfidence

HTH

PR #277, opened 2017-04-05
Added comment.

crisguitar:
Thanks for the post, I tried to do the same for a class attribute (AttributeTargets.Class) and I am getting a null object every time I get the custom attributes. Does this only work for Methods? Or how can I make it work with classes? Thanks.


ploeh:
Thank you for writing. I'm currently travelling, but I'll respond when I have time.

ploeh:
Sorry about the delay, but I've now left a response on the page: http://blog.ploeh.dk/2014/06/13/passive-attributes

HTH

PR #279, opened 2017-04-12
Comment on FizzBuzz kata in F#: stage 1

ajukraine:


ploeh:
Thanks for writing. Running the branch on my laptop installation of my blog, Jekyll complains about the `gist` tag:

```
  Liquid Exception: Liquid syntax error: Unknown tag 'gist' in C:/Users/mark/Documents/ploeh/Blog/_posts/2012-07-20-FizzBuzzkatainF#stage1.html
```

I'm hardly a Jekyll or Ruby expert, so perhaps I'm doing something wrong. If so, please advise.

ajukraine:
Hm. It's 3rd party tag - https://github.com/jekyll/jekyll-gist
If you don't want to install it, I'll alter my comment with inline code

ploeh:
For security reasons, GitHub Pages doesn't allow user-installed tags, so it's not a question of whether or not I want to install it...

IIRC, GitHub Pages may include a restricted set of third-party tags, but I don't know if that one is one of them. Do you?

ploeh:
Closing due to inactivity.

PR #282, opened 2017-04-26
Added comment about holiday calculations

aklefdal:
The view is not tested. I'm betting on you testing that the comment looks OK :-)


ploeh:
Thank you for writing. I've left a response on the page: http://blog.ploeh.dk/2017/04/24/simple-holidays

PR #287, opened 2017-06-29
Comment on Haskell time free monad post

benjamin-hodgson:


ploeh:
Thank you for writing 👍 😄  I've left a response on the page: http://blog.ploeh.dk/2017/06/28/pure-times-in-haskell

PR #289, opened 2017-06-29
answered your question

benjamin-hodgson:


ploeh:
Thank you for writing; I left a response on the page: http://blog.ploeh.dk/2017/06/28/pure-times-in-haskell

PR #293, opened 2017-07-17
Reader comment

Kurren123:


ploeh:
Thank you for writing. I've left a response on the page: http://blog.ploeh.dk/2017/01/30/partial-application-is-dependency-injection

HTH

PR #295, opened 2017-07-25
Blog comment on 'Simple holidays'

ThomasBouman:


ploeh:
Thank you for writing 👍 I left a short response on the page: http://blog.ploeh.dk/2017/04/24/simple-holidays

PR #296, opened 2017-07-27
Update 2017-07-24-combining-free-monads-in-haskell.html

Porges:
Add a comment — an example of using the `Sum` of both functors instead of `FreeT` over both.

I hope I haven't made any HTML mistakes 😁


ploeh:
Thank you for writing 👍 I left a response on the page: http://blog.ploeh.dk/2017/07/24/combining-free-monads-in-haskell

Porges:
Yes, I don't think it would be very useful in F# since you can only write code against concrete functors in the first place. (A type provider for free monads might work, but requires a lot more effort!)

PR #305, opened 2017-08-21
Comment about Fody.With

mikhailshilkov:


ploeh:
Thank you for your comment 👍

mikhailshilkov:
The idea with PRs for comments is OK, but the experience of direct HTML editing is quite tricky. Especially, lack of live preview (I can see my typos now, after you merged it) and lack of formatting (e.g. code formatting). Would be easier if it was something like markdown. Also, a direct link from the blog to the exact file in github would help.

ploeh:
I'm aware that the system's hardly ideal, but it balances various concerns, as outlined [here](http://blog.ploeh.dk/2013/03/03/moving-the-blog-to-jekyll).

I'm perfectly willing to entertain concrete suggestions for improvements.

PR #306, opened 2017-08-21
Leveraging Impromptu for UT Arrange phase

Aszan:


ploeh:
Thank you for your comment. I've left a response on the page: http://blog.ploeh.dk/2017/08/15/test-data-builders-in-c

PR #308, opened 2017-08-22
Impromptu Interface within UT details

Aszan:


ploeh:
Thank you for writing. I left another response on the page: http://blog.ploeh.dk/2017/08/15/test-data-builders-in-c

PR #309, opened 2017-08-22
Partial Impromptu Interface

Aszan:


ploeh:
Now I think I get it! Thank you for taking the time to explain 👍

PR #310, opened 2017-08-23
Comment : Thanks for clarification

harshdeep21:
Comment : Thanks for clarification

### DIFF
--- a/_posts/2017-08-21-generalised-test-data-builder.html
+++ b/_posts/2017-08-21-generalised-test-data-builder.html
@@ -500,4 +500,11 @@ tags: [Unit Testing]
 		</div>
 		<div class="comment-date">2017-08-22 13:27 UTC</div>
 	</div>
+	<div class="comment">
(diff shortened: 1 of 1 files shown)


ploeh:
Glad to be of use 😄 

BTW, thank you for asking that question 👍 I think that I still have many C# readers, and there's likely many other readers who thought as you did. It takes courage to ask the 'stupid questions', but it's _so_ important that someone does that, because it gives many other people a chance to learn, even when they're afraid to ask.

And just to be clear: I don't really think it was a stupid question. There aren't any stupid questions. It's impossible to know everything. Just because one doesn't know a particular thing doesn't mean that one is ignorant 😄

harshdeep21:
Just to echo your point. I come from "[No such thing as a stupid question.](https://en.wikipedia.org/wiki/No_such_thing_as_a_stupid_question#There_are_no_stupid_questions)" league. 👍 😆 

Thanks as always for sharing your thoughts.

PR #312, opened 2017-08-29
Comment on "The Builder functor"

andresmoschini:


ploeh:
Thank you for writing 👍 I've left a response on the page: http://blog.ploeh.dk/2017/08/28/the-builder-functor

PR #314, opened 2017-09-01
Comment on WebAPI DI constructor vs Application_start.

andyg8180:
curious about the container being registered in the constructor since WebApplication's constructor is called multipel times on startup and on each request.  thank you!


ploeh:
Thank you for writing 👍  I've left a response on the page: http://blog.ploeh.dk/2012/10/03/DependencyInjectioninASP.NETWebAPIwithCastleWindsor

PR #315, opened 2017-09-12
A comment on Test Data Builder.

otto-gebb:


ploeh:
Thank you for the link to that _Software Engineering_ question 👍 I added an answer to your question - not so much for your benefit, but for the benefit of anyone else who might encounter that question in the future.

PR #323, opened 2017-10-06
Correct term for 'opposite' in monoid,semigroup article

davidroberts63:
converse: Engage in conversation
inverse: opposite, contrary


ploeh:
Thank you 👍 

AFAICT, _converse_ can _also_ mean opposite or inverse - at least [according to dictionary.com](http://www.dictionary.com/browse/converse). How reputable that site is, I don't know, but it has served me well in the past.

Still, no reason to use a word with ambiguous meaning when a clearer term exists, so thank you 😄

PR #324, opened 2017-10-06
Update 2017-10-06-monoids.html

khaledh:


ploeh:
Thank you 👍 I don't think that I have much to add 😄

PR #326, opened 2017-10-08
Possible Typo

awright18:
referring to Kent Beck's notion of Monoids, "I don't think he ever explicitly calls it" I think should be "..calls it out" or "says it"


ploeh:
Thank you 👍

PR #327, opened 2017-10-08
typo "applicably" should be "applicable"

awright18:
"A functor is a simple, but widely applicably abstraction." should be "A functor is a simple, but widely applicable abstraction."


ploeh:
Thank you 👍

PR #329, opened 2017-10-13
Add comment to "CQS versus server generated IDs"

irfancharania:


ploeh:
Thank you for writing; I'll see if I can write a response tomorrow, but I don't think I have time today...

ploeh:
I've left a response on the page: http://blog.ploeh.dk/2014/08/11/cqs-versus-server-generated-ids

HTH

irfancharania:
Thank you for the clarification ☺️

PR #333, opened 2017-10-20
add comment

baic989:


ploeh:
Thank you for writing 👍 I've left a response on the page: http://blog.ploeh.dk/2017/10/16/money-monoid

PR #334, opened 2017-10-23
Update 2017-10-23-convex-hull-monoid.html

mikhailshilkov:
Added a comment


ploeh:
Thank you for writing 👍 It could take me a day or two before I have time to answer.

ploeh:
I've left a response on the page: http://blog.ploeh.dk/2017/10/23/convex-hull-monoid

HTH

PR #336, opened 2017-10-24
I think this 'H' needs to be capitalized for the inline code thus far…

ardave:
… to compile.


ploeh:
Thank you for writing.

The code is copied from real source code, more specifically from [here](https://github.com/ploeh/Hull/blob/e53188fd9c1f58d3f18dbcacb066d1613625f33c/Geometry/Geometry.fs#L58). I'm _fairly_ certain that it's correct as is, but perhaps I'm missing something?

ardave:
It looks like you'd have to include the lower cased **hull** function, almost the entirety of the Geometry.fs file, for the blog-inlined function to compile as-is.  No doubt that would lead to poor blog readability.

It also makes more sense as-is when I read through to your linked blog post from 2015, "Visual Value Verification."

ploeh:
I admit that perhaps I could have done a better job of presenting the code in this blog post, but fixing that would be, I believe, a more fundamental change of the article.

Thank you for pointing out the issue, though 👍

PR #338, opened 2017-11-15
Add missing binary operation from example monoid

idg10:
I puzzled for a while over the original sentence, "Methods that return the same type of value as their input form monoids" because my initial interpretation was that you were considering methods with two arguments and that those would constitute the monoid's binary operator. Since monoids require two things - a set and a binary operation over that set - and the sentence in question talks about two things - values of a particular type, and methods that operate on them - that seemed like the natural way to interpret the sentence.

However, obviously not all such (methods + value sets) form monoids. So I eventually guessed that what you actually meant is that the methods are the monoid's set, and that you've left it as an exercise for the reader to work out what operator(s) you had in mind. Possibly you meant to set this very challenge in the opening paragraph, but I think it's easier to follow if specify the binary operator explicitly.


ploeh:
Makes sense. Thank you 👍

PR #340, opened 2017-11-23
Update 2017-10-10-strings-lists-and-sequences-as-a-monoid.html

giacomociti:
added a comment about free structures


ploeh:
That's a wonderful explanation 👍 Thank you very much for it 😄

giacomociti:
I'm glad you like it. Anyway I'm not a mathematician, just a regular
developer (with poor HTML skills since those <pre> tags I put in the
comment look weird). I simply used to like theory as a CS student, and I
really appreciate your writing skills and your effort to explain scary
stuff in informal yet useful terms. Thanks.

PR #341, opened 2017-11-25
Update default.html

leonpw:
change url to: https://jekyll-bootstrap-3.github.io/ The old url http://jekyllbootstrap.com/ cannot be reached.


ploeh:
Thank you 👍

PR #343, opened 2017-12-03
Added comment

vitrun:
clarify the identity for all and any.


ploeh:
Thank you for writing 👍 I've left a response on the page: http://blog.ploeh.dk/2017/10/06/monoids

PR #345, opened 2017-12-05
Added comment to blog post

howiecamp:


ploeh:
Thank you for writing 👍 I've left a response on the page http://blog.ploeh.dk/2014/05/16/configuring-azure-web-jobs

HTH

PR #353, opened 2018-01-13
Comment to the "Software design isomorphisms" article

srogovtsev:


ploeh:
Thank you for writing; I've left a response on the page: http://blog.ploeh.dk/2018/01/08/software-design-isomorphisms

PR #355, opened 2018-01-22
Update 2019-01-15-unit-isomorphisms.html

britebit:
Fixed typo. I think rhetorical is more correct in this case.
The line ending change was introduced by the editor on GitHub. Feel free to ignore that.


ploeh:
Thank you 👍

PR #357, opened 2018-01-28
Added comment.

marlemie:


ploeh:
Thank you for writing; I left a (short) response on the page: http://blog.ploeh.dk/2017/10/30/tuple-monoids

PR #363, opened 2018-03-05
Added a comment to "Abstract class isomorphism"

MaxKot:


ploeh:
Thanks for writing 👍 I'll write a response, but it may take me a couple of days...

ploeh:
Thank you for writing; I've now left a response on the page: http://blog.ploeh.dk/2018/02/19/abstract-class-isomorphism#13cc2c9a28114b52add8942d3ac3c94d

PR #366, opened 2018-03-08
Comment to Abstract class isomorphism

ciprian-vilcan:
Added a comment about language agnosticism


ploeh:
Thank you for writing 👍 I've added a response to the page: http://blog.ploeh.dk/2018/02/19/abstract-class-isomorphism#7aac063a684047ae9561c6666a42f908

PR #368, opened 2018-03-20
Comment on property code smell article.

julealgon:
Added a simple comment on the following article:
http://blog.ploeh.dk/2011/05/26/CodeSmellAutomaticProperty/

Parts of the article would do well with an update considering new language features like readonly automatic properties.


ploeh:
Thank you for writing 👍 I've left [a response on the page](http://blog.ploeh.dk/2011/05/26/CodeSmellAutomaticProperty#389e594137564a3bb5c9547807db3614).

PR #369, opened 2018-03-22
Added a comment to "Abstract class isomorphism".

MaxKot:
**Work in progress**
What code highlighter do you use? I have some code examples in the comment but I'm not sure how to add proper highlighting so that it matches the highlighting style of your code.

### DIFF
--- a/_posts/2018-02-19-abstract-class-isomorphism.html
+++ b/_posts/2018-02-19-abstract-class-isomorphism.html
@@ -696,4 +696,75 @@ game.CreateMaze(<span style="color:blue;">new</span>&nbsp;<span style="color:#2b
 		</div>
 		<div class="comment-date">2018-03-11 9:35 UTC</div>
 	</div>
+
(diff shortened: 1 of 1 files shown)


ploeh:
You don't _have_ to colour your code samples; a simple `<pre>` will do in a pinch. At least, _I_ don't require people to colour their code sample while commenting. I'm just happy that someone cares enough to comment at all.

In order to answer your question, though, I use the _Copy as HTML_ extension for Visual Studio. In other words, code highlighting on my blog isn't done by a post-processing tool, but is part of the source.

MaxKot:
Thanks for the quick reply! Now the comment looks more readable and should be ready for merging.

ploeh:
Thank you for writing. I've left a [response on the page](http://blog.ploeh.dk/2018/02/19/abstract-class-isomorphism#2e3ab0c8da334d5cb066eab2f492c50e).

PR #370, opened 2018-04-03
Added a comment to "Abstract class isomorphism".

MaxKot:


ploeh:
Thanks for writing 👍 It's a great comment, and I don't think that I've _that_ much to say in response, because I largely agree, but I'd still like to add a comment or two. I am, however, currently travelling and doing paid work for a customer, so it may take some days before I get around to it.

PR #372, opened 2018-04-13
Add comment about FX and decimal fractions

idg10:


ploeh:
Thank you for writing 👍 I've added [a response to the page](http://blog.ploeh.dk/2017/10/16/money-monoid/#2488dac03ee64cd1844d2a4a88374c26).

PR #381, opened 2018-05-14
Update 2017-08-07-f-free-monad-recipe.html

ymassad:
An example


ploeh:
Thank you for writing; I've left [a response on the page](http://blog.ploeh.dk/2017/08/07/f-free-monad-recipe/#57af39a3aff045feaeac211222d6e365).

HTH

PR #384, opened 2018-05-26
Add comment about pattern synonyms

benjamin-hodgson:


ploeh:
Thank you for writing 👍 I've added [a response, and a follow-up question](http://blog.ploeh.dk/2018/05/14/project-arbitraries-with-view-patterns/#85761f5529fd4806a4e4202ed4b319c5), to the page.

PR #385, opened 2018-05-27
Answered your questions

benjamin-hodgson:
I also fixed the type signature in my last comment :)


ploeh:
Thank you for elaborating. That all makes sense to me.

PR #395, opened 2018-07-04
Add missing word

moodmosaic:
At *"Second, there seems [?] some calculations"* it looks like something is missing.

---

Great article, by the way. Looking at this:

```haskell
tryAccept :: Int -> Reservation -> MaybeT ReservationsProgram Int
```

I wonder why the full source code isn't somewhere online (GitHub, GitHub Gist). Or it already is? Myself, but perhaps also others, could have a look at the whole source code.


moodmosaic:
>but I'll need to polish it a bit before I can do that

*Might* be able to help, if you want, (until end of August).

ploeh:
Thanks, but I'm pretty sure it's not what you think it is. Most of the code is C#. The Haskell code is just a little translation, in order to show how much better Haskell is at certain things. (There's also a nice F# translation.)

moodmosaic:
>The source code is currently not available, but I'll post it later.

Just saw it, right [there](https://github.com/ploeh/dependency-injection-revisited/blob/b6518e6992ac9f64a6976abccce32c79b5d6a2f2/Haskell/src/BookingApi.hs#L47)! 🎉

ploeh:
Yes, I meant to tell you, but at least, I have added a link to the repository from the article itself.

PR #396, opened 2018-07-09
Removed duplicate words

moodmosaic:
Not sure if those are duplicates, but I thought I should create a pull request anyway.


ploeh:
Thank you for the pull request. The text as written is, however, intentional. 

[James while John had had had had had had had had had had had a better effect on the teacher](https://en.m.wikipedia.org/wiki/James_while_John_had_had_had_had_had_had_had_had_had_had_had_a_better_effect_on_the_teacher)

moodmosaic:
o.O

PR #407, opened 2018-10-04
Added comment

Overlord-Zurg:
This is like my second pull request ever, so please let me know if I borked the process. Or if my question is dumb. Or if the html is improperly formatted. Or if I'm being too neurotic.


ploeh:
Thank you for writing; it looks fine 👍 

I'll respond as soon as I have time, but it may take a few days.

ploeh:
FYI, I've now added a [response to the page](http://blog.ploeh.dk/2015/11/30/to-log-or-not-to-log/#e04ac5169abd4b58a1e5dd0ac42878ca).

PR #409, opened 2018-10-18
comment in article on applicative password list about currying concat

bender2k14:


ploeh:
Thank you for writing; I've left a [short response on the page](http://blog.ploeh.dk/2018/10/15/an-applicative-password-list/#b4e76681ea894aa3be1e6b836343c148).

It's basically just a slightly expanded thank-you response, but I wanted to leave it on the page itself, so that other readers who come by later have a chance to infer that I agree with you.

bender2k14:
You are very welcome.  Helping the readers like that is a good idea.

But really, _thank you_ for writing this blog and especially these series within series about software design and category theory.  These are topics that I want to understand much better, and your articles are at just the right depth for me right now.

PR #410, opened 2018-11-06
comment in article on maybe applicative functor about currying in C#

bender2k14:
I added a comment about currying in C# to your "The Maybe applicative functor" article.  I hope you don't think I am being too nit picky by pointing out the same thing as I did before but this time on another article.


ploeh:
Thank you for writing; I've left [a response on the page](http://blog.ploeh.dk/2018/10/29/the-maybe-applicative-functor/#0bb16509bbea463bab48c9e97acea2c5).

PR #434, opened 2019-02-05
Update 2019-02-04-how-to-get-the-value-out-of-the-monad.html

SeanGDonohue:
Added comment on combing two options and Cat function.


ploeh:
Thank you for writing 👍  I've left [a response on the page](http://blog.ploeh.dk/2019/02/04/how-to-get-the-value-out-of-the-monad/#f8d3e120bf9a46d09b479106812e7020).

PR #435, opened 2019-02-07
Comment 2019-02-04-how-to-get-the-value-out-of-the-monad.html

AchimStuy:


ploeh:
Thank you for writing. I've left [a response on the page](http://blog.ploeh.dk/2019/02/04/how-to-get-the-value-out-of-the-monad/#885dd7685e8142d7b5af72796ea8c770).

PR #436, opened 2019-02-07
Ask question about composition with async methods in the chain

ralphhendriks:


ploeh:
Thank you for writing. Stay tuned; next week's blog post should, hopefully, answer your question 😄

ploeh:
FWIW, I've now added [a short reply on the page](http://blog.ploeh.dk/2019/02/04/how-to-get-the-value-out-of-the-monad/#3588d6472d4f4e76832925221aae374a).

PR #451, opened 2019-02-15
comment on 2019-02-11-asynchronous-injection.html

Thaoden:


ploeh:
> All in all, I now think a bit differently about the controller responsibilities: Being at the boundary of the system, it is arguably the best place to make calls to external systems. If and how the information gathered from the outside _is used_ however is still up to the business objects. Thanks, Mark, for the insight!

Oh, neat 😊 You're welcome 👍

Issue #389, opened 2018-06-06
More tags


ploeh:
[The tag cloud](http://blog.ploeh.dk/tags) on the blog is becoming increasingly useless, since everything I write today basically just gets shoved into the _Software Design_ tag, which, at the time I'm writing this, has 172 entries. This is clearly not an aid to readers who try to navigate the blog.

While I'll keep the tags that are already present, I'd like to add some more tags that could offer more granular navigation. The intent is to retrofit such tags unto existing articles.

I can't figure out which tags to create, though!

With this issue, I hereby solicit suggestions.

New tags ought to cover more than a pair of articles. I admit that I've made some mistakes when I originally created some of the tags. For example, I thought I'd be writing more on _WPF_ than I ultimately did; that tag only has two entries, but at least that's better than the _Sync_ and _Albedo_ tags.

New tags should have more entries than that (at least 5-10). Additionally, any new tags should be distributed across articles that aren't otherwise natural clusters. For example, one could easily invent a _Monoid_ tag and add it to all the articles in the [Monoids](http://blog.ploeh.dk/2017/10/06/monoids) article series, but if that tag doesn't apply outside of that article series, then what value would it add? That's the reason I haven't already added _Monoid_, _Functor_, or _Category Theory_ tags.

If you have an idea for one or more tags that could be added to the blog, please add your suggestion below, preferably with some suggested articles that'd belong to the new tag.

Better yet, send your suggestion as a pull request 😉

adamchester:
Why do you increasingly keep adding to just the “Softward design” tag?

ploeh:
That's my issue! Which other tag(s) do you think I should use for my recent articles?

ploeh:
I've now added a [Design Patterns](http://blog.ploeh.dk/tags/#Design%20Patterns-ref) tag. It spans articles from 2010 to 2018, and I have two more articles with that tag queued up for publication in 2019. It already includes 26 articles, so I think it fits all the above criteria.

PraveenTalari:
Could you please add new header called "Series" to
http://blog.ploeh.dk/tags/

with links to series such as
http://blog.ploeh.dk/2017/10/04/from-design-patterns-to-category-theory/

BTW, do you have any other series of articles as such?

ploeh:
[...]

PR #454, opened 2019-02-18
Added Article Series tag

ploeh:
This is based on a discussion had at
https://github.com/ploeh/ploeh.github.com/issues/389#issuecomment-463201337

In short, the purpose of the tag is to identify the entry points to
various article series.


PraveenTalari:
looks good :thumbsup:

PR #463, opened 2019-03-20
Comment to "Test data without Builders"

bigdnf:


ploeh:
Thank you for writing 👍 I've left [a short response on the page](https://blog.ploeh.dk/2017/09/11/test-data-without-builders/#f657b7db01d24d6cbc1b358ed6a13a20):

PR #466, opened 2019-04-05
ladeak comment for An example of state-based testing in C#

ladeak:
Question for An example of state-based testing in C#


ploeh:
Thank you for writing. I've left [a response on the page](https://blog.ploeh.dk/2019/04/01/an-example-of-state-based-testing-in-c/#39fa5e0bfdf24153854a260addf32d0e).

PR #468, opened 2019-04-12
The embedded world nightmare... 

zen-coder:
I agree with you all along and I would like to add a striking example from the embedded world... You can actually generate test cases to reach 100% Statement/Branch or even MCDC Coverage! 

https://www.qa-systems.com/tools/cantata/autotest/
https://en.wikipedia.org/wiki/Modified_condition/decision_coverage

Basically it can be useful to characterize the current state of you software and use it for non regression. Unfortunately it is often misused. 
If I have a function "Add(A,B){return A-B;}" I will still have it covered and have all necessary assertion (pre/post conditions). At best it only proves the code will not crash on most paths... And you get a Valid ISO26262 certified report of 100% coverage from suppliers!

Often, in avionics and automotive, each test case should have a "6 eyes confirmation" (meaning two reviews) and be traceable back to a requirement. But it is often extremely hard to achieve 100% coverage with these. Typically, some header only libraries may have functions you do not need or some "catch(...)" clause might be unreachable. Many kernel level thing are heavily timing dependent. etc...

Also some static analysis requirements may drive one nuts, like the MISRA compulsory "default" in a switch: https://www.misra.org.uk/LinkClick.aspx?fileticket=oIhIwlbUZz0%3D&tabid=57 

These often force you to write an unreachable switch default statement...  Or to treat one of your normal case as default to reach 100%!

I could go for hours on that. We are forced by the ISO26262 to reach 100% coverage! I am happy not to be the only one to think what you wrote! Cheers.

### DIFF
--- a/_posts/2015-11-16-code-coverage-is-a-useless-target-measure.html
+++ b/_posts/2015-11-16-code-coverage-is-a-useless-target-measure.html
@@ -201,4 +201,4 @@ tags: [Productivity, Unit Testing]
 	<p>
 		On the other hand, it can be useful to measure code coverage once in a while, but it shouldn't be your main source of information about the status of your source code.
 	</p>
-</div>
+</div> 


ploeh:
Well, yes, none of that sounds surprising 😄

ploeh:
I'm going to close this pulle request unmerged, since there's really nothing to merge...

PR #469, opened 2019-05-13
corrected second for first typo

bender2k14:


ploeh:
Thank you, but I'm _fairly_ sure that the sentence as written correctly describes what's going on..(?) 😄

ploeh:
Correct me if I'm wrong, though 😄

bender2k14:
Oh, yes.  You are correct.  Sorry for my mistake.

ploeh:
FWIW, once you brought this to my attention, I had to read the sentence a couple of times before I came to that conclusion. I could probably have phrased it better...

bender2k14:
I know this operation as the tensor product of a two-dimensional column vector with a three-dimensional row vector
https://en.wikipedia.org/wiki/Tensor_product

ploeh:
Uh, oh, it's been 25+ years since I had to think about vectors... 😓

PR #471, opened 2019-05-13
comment/question about functor containing monoidal values still forming a monoid

bender2k14:


ploeh:
Thank you for writing 👍 I've left a [response on the page](https://blog.ploeh.dk/2019/04/22/applicative-monoids/#a164909adb884cd78b309a81029a2dd8).

PR #472, opened 2019-05-15
added discussion/question about bifuctor vs functor

bender2k14:


ploeh:
Thank you for writing 👍 I've added [a response to the page](https://blog.ploeh.dk/2019/01/07/either-bifunctor/#19253916a1bd49a08b19b15898816657).

PR #473, opened 2019-05-30
Applicative tuple string int comment

bender2k14:


ploeh:
Thank you for writing; I've left [a response on the page](https://blog.ploeh.dk/2019/04/22/applicative-monoids/#daa6444c8fb0484a8bf1accdb3cc544b).

PR #475, opened 2019-05-31
asking about definition of pure for Tuple<string, B>

bender2k14:


ploeh:
Thank you for writing. I've left [a response on the page](https://blog.ploeh.dk/2019/04/22/applicative-monoids/#ce69d4440f314827a8ec441511d0ce71).

PR #483, opened 2019-07-12
added comment about reduce and fold

bender2k14:


ploeh:
Thank you for writing; I've left [a response on the page](https://blog.ploeh.dk/2019/04/29/catamorphisms/#43e48df0645b4c05992b0f599cc71d10).

PR #486, opened 2019-07-31
Comment about async monads

bigdnf:


ploeh:
Thank you for writing; I've left [a comment on the page](https://blog.ploeh.dk/2019/02/04/how-to-get-the-value-out-of-the-monad/#db0c16dd403d43f0947b536dc91ec1b1).

PR #489, opened 2019-08-04
added comment about how to best create a desktop application in F#

bender2k14:


ploeh:
Thank you for writing; I've left [a response on the page](https://blog.ploeh.dk/2014/01/22/a-functional-architecture-with-f/#66d2b843a43341ad821f17e18d426b52).

PR #490, opened 2019-08-04
added comment asking Mark how he does NuGet now

bender2k14:


ploeh:
Thank you for writing; I've left [a response on the page](https://blog.ploeh.dk/2014/01/29/nuget-package-restore-considered-harmful/#2de063c09884401f9ef22fd1662c74c0).

PR #494, opened 2019-08-22
Question about web front end

bender2k14:


ploeh:
Thank you for writing; I've left [a response on the page](https://blog.ploeh.dk/2014/01/22/a-functional-architecture-with-f/#7830ba29e87a46eebb7aba4f456e63ea).

PR #497, opened 2019-09-12
Haskell operators

andre-silva-cardoso:
Comment about Haskell operators


andre-silva-cardoso:
Thanks for your answer. I must take some time to learn basic Haskell (at least enough for reading) J

 

Best regards,

André Cardoso

PR #499, opened 2019-09-21
comment about F# records and limit on backwards compatibility

bender2k14:


ploeh:
Thank you for writing; I've left [a comment on the page](https://blog.ploeh.dk/2017/09/11/test-data-without-builders/#cb5fa591f062499db86d428fe4a02e6c).

PR #503, opened 2019-10-03
remove rounded border from navbar

chekkan:
Use bootstrap [static top navbar](https://getbootstrap.com/docs/3.3/components/#navbar-static-top)

Screenshot 2019-10-03 at 20 39 36


ploeh:
Thank you 👍 This has always slightly bothered me, but never enough to actually fix it (since I didn't know how to do it).

PR #504, opened 2019-10-07
Fix incorrect 'next' link on page 2

### LINKED ISSUE #246: Incorrect 'next' link on page 2
# Steps to reproduce
1. Go to http://blog.ploeh.dk/page2
2. Scroll down to the bottom
3. Click _Next_
# Expected behaviour

The link takes you back to the 'first' page, which is, implicitly, http://blog.ploeh.dk
# Actual behaviour

The link points to http://blog.ploeh.dk/page1, which doesn't exist, resulting in a _404 Not Found_ result.

---

IIRC, this once worked, but I may be wrong about that. If, on the other hand, I remember correctly, this bug may have been introduce by the infamous Jekyll update some time ago.

Anras573:
Hi @ploeh 

I read your blog from time to time, and have fixed the incorrect 'next' link on page 2

Fixes #246

### DIFF
--- a/index.html
+++ b/index.html
@@ -19,11 +19,11 @@ description: "Danish software design"
 <!-- Pagination links -->
 <div class="pagination hidden-print">
   {% if paginator.previous_page %}
-    <a href="/page{{paginator.previous_page}}" class="next">Next</a>
+    <a href="{{ paginator.previous_page_path }}" class="next">Next</a>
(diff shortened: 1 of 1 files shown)


ploeh:
Oh, wow, you just closed a 3+ year old issue 😲 Thank you 💯

PR #508, opened 2019-10-25
comment about testing functions with a known complexity

bender2k14:


ploeh:
Thank you for writing. I've left [a response on the page](https://blog.ploeh.dk/2019/10/07/devils-advocate/#cb15452b8c96429998efd50b67373da3).

PR #512, opened 2019-11-06
Add comment

cmeeren:
Please let me know if anything is unclear (or if you spot any typos), and I'll update the comment.


ploeh:
Thank you for the great example 👍 It may take me some days before I can concoct an appropriate answer, but I'll post a link here once I've done that.

ploeh:
I was hoping to address this the next couple of days, but looking over your sample code, I've noticed that all the return values are just given as `Async`. I suppose that they're `Async<ProofId>`, `Async<bool>`, `Async<unit>`, and so on... I *think* that I can infer them all, but it'd be a help if you could update the example code..(?)

cmeeren:
Hm, they are correct in the source, so apparently things like `Async<ProofId>` needs escaping of `<` and `>`.

ploeh:
Sorry, I should have looked at the source 😳 

I can do that 👍

ploeh:
This turned out to be an instructive exercise, so I'll write my response as a blog post. I've reserved December 2 in the [F# Advent Calendar](https://sergeytihon.com/2019/11/05/f-advent-calendar-in-english-2019) for the post.

cmeeren:
Great, I'm very much looking forward to it! 👍 

By the way, I have come across another case that may or may not be useful for that blog post.

As I said in my comment, I have an API for allowing users to register. Le me briefly share some context first. The registration is done in three steps:

1. First a registration resource is created. This is often done by the business ("inviting" a user), which can set certain privileged fields like email, username, customer number, roles, etc. that can't be edited by the end user.

2. The end user then fills out other information (name, phone, etc.) by PATCH-ing other fields on this resource (using the front-end, of course)

3. When the registration is in a valid state and can be submitted, the user can submit it (POST to /registrations/{id}/complete), which will actually create the user using the information in the registration.

End of background information.

Now, registering users naturally requires checking whether a user already exists with the specified username. There's a constraint in the DB to ensure duplicates never happen, but we also check this explicitly at the beginning, when creating a registration, so that a useful error is returned at the start if the user exists.

Ideally, this check would be performed in the domain layer, in the function that creates a registration:

```f#
createRegistration:
  Email -> Role list -> (*...*) -> Result<Registration, CreateRegistrationError>
```

Placing the check here is desired among other reasons because, currently, there is a rule that the username must be set to the email address (hence `createRegistration` accepts `Email`, but not `Username`). This knowledge is owned by this function (and only indicated to the outside world due to `Username` not being part of the signature). Since the knowledge of what the username actually is doesn't exist anywhere else (before creating the registration), this function should also be where the username check is performed.

I see three ways to solve this:

1. Pure, testable without mocks: Have `createRegistration` accept a list of all usernames. This is fine in our case, because we don't have that many users in this system. But it won't scale for millions of users. The composition function that calls `createRegistration` must then get the usernames from the DB and pass them to `createRegistration`. The benefit is that all business logic exists in pure functions. (I am not counting "get and pass all usernames" as business logic, because that's trivial. Feel free to disagree, though.)

2. Impure, requires integration tests: Have the composition function do the check after creating the registration, using the `Registration.Username` field. The composition function creates the registration, and then calls the DB to check for the existence of the particular username that `createRegistration` decided to use. The downside of this is that now we have business logic in a composition function, which requires integration tests.

3. Pure-ish, testable with mocks: Create a workflow function:

   ```f#
   registerUser:
    (checkUserExists: Username -> Async<bool>)
    (registration: Registration)  // ???
    : Async<Result<Registration, CreateRegistrationError>>
   ```

   The composition function simply calls this with the relevant arguments. Benefit: Unit testable (though not as simple as purely data-in-data-out functions). Downside: The function must either accept a registration as pure data as shown above, with the composition function creating the registration (or it can accept a thunk to create a registration, which is essentially the same), which essentially renders it trivial; or it must take a parameter `creatRegistration` with a signature that matches the domain function signature, which means the workflow must then accept all relevant parameters too, which makes it harder to test; or it can directly call `createRegistration`, which also makes it harder to test since then you're testing that, too (and it again requires all the parameters of the domain function).

As you can see, this is another example where sandwiching is not trivial and creates some problems. I have come across very similar variants several times.

ploeh:
I'll refrain from extensive commenting until you've seen the article. It took me most of the week to write it, but it has a lot of detail. I hope that it answers some of your questions. If it doesn't, I'd rather have the discussion where more people can see it, but let's see how it unfolds after December 2.

I'd like to comment on one thing you wrote here, since I consider it independent of functional programming.

> Have createRegistration accept a list of all usernames. This is fine in our case, because we don't have that many users in this system. But it won't scale for millions of users.

Why not?

Let's do the maths. As I understood you, user names are email addresses? (If not, I assume that they're shorter, so that just strengthen my position.) Let's assume that the average email address is 25 characters. In .NET, strings are always Unicode, so that's 50 bytes. A million of those is approximately 50 Mb, even before you decide to put them into any sort of fancy data structure. That easily fits in RAM.

People often make such throwaway comments. After all, _a million_ sounds like a lot, but in terms of modern memory capacity, it isn't.

Would I go with that alternative, then?

Probably not, but for another reason. System-wide constraints such as _the user name must be unique_ can't be addresses with code. You'll need some kind of architecture to enforce a rule like that.

Even if you write a function that enforces such a rule, you'll still need to set up the architecture so that the function doesn't run concurrently.

There's plenty of ways you can do that, but I'd like to postpone that discussion until you've seen the article I've already written.

cmeeren:
Thanks! Looking forward to it.

ploeh:
FWIW, I've now published [the article](https://blog.ploeh.dk/2019/12/02/refactoring-registration-flow-to-functional-architecture).

ploeh:
BTW, I haven't forgotten that you had some other questions here, but please read the article and see how many you have left. I'll be happy to answer any other questions to the best of my ability.

cmeeren:
Thanks! I appreciate your taking the time to reply with a lengthy post. I'll read it as soon as I can, reflect on it, and let you know if I have further questions.

cmeeren:
I see that I have a typo in the flowchart. The final step (lower right) should be "complete registration", not "complete proof". Do you think it's confusing enough to warrant a correction? I don't have the original, so it'll have to be a note in the text, or an edit to the png with a strikeout/insertion etc. (Or I could make a new one, but I'd rather avoid it.)

ploeh:
I didn't notice; I wonder if anyone else will. I mostly used the code as a the 'source of truth'.

If there's both code and a picture, the picture lies, because the code is what gets executed 😄

PR #513, opened 2019-11-18
added comment about additional stack trace information

bender2k14:


ploeh:
Thank you for writing. I've left [a response on the page](https://blog.ploeh.dk/2019/11/18/small-methods-are-easy-to-troubleshoot/#8a66b203db9245d8b3db5e40399f039e).

PR #514, opened 2019-11-18
added comment about pdb files

bender2k14:


laurencee:
Yep was going to write this same comment, that was extremely surprising to me to find a .NET vet wasn't aware this was one of the major purposes of packaging your pdb files with release dlls.

bender2k14:
@laurencee, do you always include your pdb files with your release DLLs (or does it depend on something)?

laurencee:
For the past 10 years every .NET app i've worked on I've built in release mode and included the pdbs with release dlls/exes for deployments/releases. It doesn't depend on anything else.

This includes Microsoft C++ projects actually, they basically work the same way.

It's actually more flexible than having to package them with release binaries. You dont technically need the pdbs alongside the dlls at the time the error occurred as long as you record the managed exception with the full details/offset information, you can later pair that with the pdbs that are stored elsewhere and retrieve the line numbers. Not many people are aware of that AFAIK.

Just as a very specific example from my past, a real reason you might not want pdbs is if you were proving "deterministic builds" i.e. delivering a set of binaries that could be reproduced byte for byte from the same source code, pdbs didn't use to be capable of being created deterministically so they'd had to be excluded from the process. 

Outside of that I don't know of any real reason why you wouldn't package the pdbs them along with the release binaries.

bender2k14:
> ...I don't know of any real reason...why you wouldn't package the pdbs them along with the release binaries.

Is there any (noticeable) impact on performance?

laurencee:
They have no impact on performance outside of a stacktrace at which point the difference doesn't matter anymore since you've already paid a massive cost throwing an exception.

ploeh:
Thank you for writing. I've left [a response on the page](https://blog.ploeh.dk/2019/11/18/small-methods-are-easy-to-troubleshoot/#2818246690a94b7b9c81ddfb47fc6d89).

PR #521, opened 2019-12-07
added an example of a pure-impure-pure-impure composition

bender2k14:
Thanks for fixing my broken link in my previous comment :)


ploeh:
Thank you for writing. I've left [a response on the page](https://blog.ploeh.dk/2019/12/02/refactoring-registration-flow-to-functional-architecture/#189069ae41704900b56403288656a8fe).

PR #523, opened 2019-12-09
A pure-impure-pure composition that I think has one responsibility

bender2k14:


ploeh:
Thank you for writing. I've left [a response on the page](https://blog.ploeh.dk/2019/12/02/refactoring-registration-flow-to-functional-architecture/#f2db5259c16e4241b53934ae4dcb17a0).

PR #526, opened 2019-12-12
Update 2019-04-01-an-example-of-state-based-testing-in-c.html

ledbutter:
If I was a blogger I probably would have turned this comment/question into a blog post in itself referencing this series, but since I'm not, here is my question/comment.


ploeh:
Thank you for writing. I've left [a response on the page](https://blog.ploeh.dk/2019/04/01/an-example-of-state-based-testing-in-c/#0c024d78ebeb418d9cc70c7a267daeed).

PR #533, opened 2019-12-16
comment about aspects of type systems

bender2k14:


ploeh:
Thanks for the comment 👍 I don't think I've anything to add, so I don't currently plan to respond more than this 😄

bender2k14:
You're welcome.  I am happy to contribute :)

And thank you for writing such a good post in the first place!

I had recently and carefully reviewed all those linked Wikipedia pages, so I was glad to find a place to share some of what I learned.  (I was hoping that my company would have a blog by now where I can share these thoughts, but it is not quite set up yet.)

PR #536, opened 2019-12-20
Add my comment about TypeScript

rdeneau:


ploeh:
Thank you for writing 👍  While you're welcome to suggest edits of all of _my_ typos and errors (after all, I'm the person reviewing the pull request), I consider other readers' comments to belong to them, typos and all. I don't think it's proper to edit them.

Could you please undo the changes to Tyson Williams' comment?

ploeh:
Now that I think about it, it may seem that [I'm not entirely consistently obeying my own rules](https://github.com/ploeh/ploeh.github.com/pull/530), but I think there's a distinction between editing markup and changing peoples' words.

I follow this principle because I want commenters to trust me to not change their words while they aren't looking.

What you've done is quite innocuous, but I hope it's better to have a clear policy than no policy.

bender2k14:
I will review the changes to me comments when I get to my laptop (since the diff view on my phone app isn't good enough for this case). 

Also, don't forget to set the comment ID to s GUID.

ploeh:
> Also, don't forget to set the comment ID to a GUID.

No worries, I can do that if need be...

ploeh:
Thank you, both 👍  Like the previous comment to this article, I don't think that I've much to add, so I currently don't plan to publish an answer.

rdeneau:
@bender2k14 : thank you. Your comment on the article is very elaborate _(so typos can easily sneak into)_ and brings value. This is why I found useful to fix (just) the typos, hoping not to add my own ones. I appreciate that you're OK, not seeing it as rude.
@ploeh : thanks for the merge.

ploeh:
@rdeneau To be clear, I don't see this as rude at all - quite the contrary 😄 

I'm just keeping to a easily definable policy to avoid any future slippery slopes.

Thank you 👍

PR #545, opened 2020-01-14
Update 2017-02-02-dependency-rejection.html

dhulke:
Question about whether dependency injection is really not needed in FP and style of code.


ploeh:
Thank you for writing 👍 I've left [a response on the page](https://blog.ploeh.dk/2017/02/02/dependency-rejection/#5bfb598be4d34755b4c2272dc1814df4).

PR #548, opened 2020-01-20
Comment on 2020-01-20-algebraic-data-types-arent-numbers-on-steroids.…

drewjcooper:
…html

Add a comment

### DIFF
--- a/_posts/2020-01-20-algebraic-data-types-arent-numbers-on-steroids.html
+++ b/_posts/2020-01-20-algebraic-data-types-arent-numbers-on-steroids.html
@@ -253,4 +253,24 @@ fromList [("GBP",500 % 1),("JPY",250 % 1),("USD",900 % 1)]</pre>
 		</div>
 		<div class="comment-date">2020-01-20 20:54 UTC</div>
 	</div>
+	<div class="comment" id="a496e710c3ea4391a9de433da3e6d54d">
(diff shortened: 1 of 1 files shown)


ploeh:
Thank you for writing 👍 I'd considered replying to Jakub's comment, but yours amply does that.

Did you mean to respond to Jakub, or is the question directed at me?

ploeh:
Well, regardless of who you directed your question at, you got [a reply from Jakub](https://blog.ploeh.dk/2020/01/20/algebraic-data-types-arent-numbers-on-steroids/#bd48b3267e014ccc9b377e983369ddca) 😄

drewjcooper:
Mark, I was intending the question for yourself. I'm not sure that the code in the original post supports the point you're trying to make in the text, but it's entirely possible that i'm missing something.

The `Trade` data type doesn't inherently prevent you from adding values of different currencies, although to solve the given problem you necessarily have to separate the currency types. This seems to me, though, to be a feature of the data processing algorithm, rather than the data type. Again, am I missing something?

ploeh:
Okay, I just wanted to be sure. FWIW I've now left [a response on the page](https://blog.ploeh.dk/2020/01/20/algebraic-data-types-arent-numbers-on-steroids/#c9069e6554934325a86a27921cff7ac3). Please write again if you don't find it satisfactory 😄

PR #567, opened 2020-02-26
comment about how the simplest overload of DateTime.TryParse is impure

bender2k14:
@ymassad, I ask you a question in this comment


ploeh:
Thank you for writing. I've left [a response on the page](https://blog.ploeh.dk/2020/02/24/discerning-and-maintaining-purity/#d2cd79f75b0c4904bd55f34887513b61).

PR #569, opened 2020-02-27
Add Elm as an example functional language

ceddlyburge:
Hi Mark

Opening Prs is the way to make comments, but I thought I might as well just propose the actual change instead.

I believe that one of your aims is to promote staticcally strongly typed functional languages, and I believe that Elm is probably the best way to do this.

Elm is targeted at front end development, and makes it easy to create websites, which is what most beginners are trying to do. The language has a strong focus on simplicity and a single recommended way of doing things, which is a refreshing change, especially in the front end world. It sits very well in it's target niche, and the runtime essentially provides an impure abstract syntax tree interpreter, leaving all your code 100% pure / functional. There is also a strong and vibrant community, which is both innovative and supportive to newcomers. It doesn't have all the features of haskell, which obviously has some good points and some bad points

Thanks

Cedd

### DIFF
--- a/_posts/2020-02-24-discerning-and-maintaining-purity.html
+++ b/_posts/2020-02-24-discerning-and-maintaining-purity.html
@@ -184,7 +184,8 @@ tags: [Functional Programming, Unit Testing, Architecture]
 		Language support <a href="#5413987a6af14316aee1b1de82aee73d" title="permalink">#</a>
 	</h3>
 	<p>
-		While no mainstream languages distinguish between pure functions and impure actions, there are languages that do. The most famous of these is <a href="https://www.haskell.org">Haskell</a>, but other examples include <a href="http://www.purescript.org">PureScript</a> and <a href="https://www.idris-lang.org">Idris</a>.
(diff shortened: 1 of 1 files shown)


bender2k14:
@ceddlyburge, can you remove the newline break that you added?

ploeh:
Cedd, thank you for writing. I'm aware of Elm, and have nothing against it. On the other hand, I've also no experience with it. I write about things that inspire me, and so far, having not worked with Elm, it hasn't had the opportunity to inspire me.

Until that changes, I'm happy to let other people write about Elm, if that's their passion.

ceddlyburge:
Hi Mark

Thanks for the reply, I appreciate it. And no worries obviously.

I remain hopeful that Elm will inspire you sometime in the Future! Haskell is not trying to be beginner friendly, and isn't. F# is a lot more approachable, especially if you come from a .net background (as I do), but as you point out in this article it isn't 100% functional. Elm is 100% functional and has a specific aim of being beginner friendly. I think it would help you on your quest :)

Cheers

Cedd

PR #572, opened 2020-03-02
fixed obverse for reverse typo

bender2k14:
Replaced "obverse" with "reverse".  I think this is what you meant.


ploeh:
What's wrong with _obverse?_

bender2k14:
Oh, I learned a new word today :)

Nevermind ;)

PR #573, opened 2020-03-06
asked if a pure function can throw an exception

bender2k14:


bender2k14:
I accidentally deleted my branch from my fork, which caused this PR to be closed. Sorry about that.  See #574 now.

PR #575, opened 2020-03-07
added comment clarifying my inquiry to the definition of a pure function

bender2k14:


ploeh:
Thank you for writing. I've left [a reply on the page](https://blog.ploeh.dk/2015/05/07/functional-design-is-intrinsically-testable/#97e4ae29436e4828813bf445ee1c37dc).

PR #577, opened 2020-03-09
comment about exception throwing being a side effect

bender2k14:


bender2k14:
Thanks for correcting my HTML bug :)

ploeh:
Thank you for writing. I've left [a response on the page](https://blog.ploeh.dk/2015/05/07/functional-design-is-intrinsically-testable/#7bdf9f1597334a40afe4b8488456d7b2).

HTH

PR #578, opened 2020-03-09
Comment about purity definition

bender2k14:


ploeh:
Thank you for writing. I've left [a response on the page](https://blog.ploeh.dk/2020/02/24/discerning-and-maintaining-purity/#a7375c806c1147c1b37187834fb04153).

PR #580, opened 2020-03-10
comment digging deeper into the definition of purity

bender2k14:
I am absolutely loving our conversations about purity.  Thank you very much joining me as I unpack this definition about which [you said](https://blog.ploeh.dk/2015/05/07/functional-design-is-intrinsically-testable/#8357b5ab8ada4f44bf17dcacd9b9b079)
> It seems to me to be a standard and non-controversial definition.

;P


ploeh:
Thank you for writing. I've left [a response on the page](https://blog.ploeh.dk/2020/02/24/discerning-and-maintaining-purity/#dd776b1b8b0d43c58171ff7fec311c2c).

PR #581, opened 2020-03-11
comment about public API design where functions have (at most) one argument

bender2k14:


ploeh:
Thank you for writing 👍 FWIW, I've left [a response on the page](https://blog.ploeh.dk/2020/03/09/polymorphic-builder/#d31b6312b4204c7e8a1c047ed49cd4e6).

I don't know if it's in any way helpful, but please write again if it isn't.

I've been behind on answering your questions, but I believe that I've now caught up. If there's a question that I forgot to answer, please let me know.

bender2k14:
> I don't know if it's in any way helpful, but please write again if it isn't.

It is helpful _and_ I will write again ;)

> I've been behind on answering your questions, but I believe that I've now caught up. If there's a question that I forgot to answer, please let me know.

That is only because I have been asking so many questions.  Given the large quantity, I think you have been very quick to respond.

I treasure the opportunity to ask you questions and value all of your responses and the time it took you to think about and write them.

PR #584, opened 2020-03-11
comment about totality

bender2k14:


ploeh:
Thank you for writing. I've left [a reply on the page](https://blog.ploeh.dk/2015/05/07/functional-design-is-intrinsically-testable/#34436c76e8274a0db7d28d70ee1f2321).

PR #587, opened 2020-03-16
comment about my motivation for studying the builder pattern

bender2k14:


bender2k14:
Sorry for my mistake with the date and reusing a previous branch name.  Thanks, as always, for correcting my time stamp :)

PR #588, opened 2020-03-17
added comment about function that randomly throws different exceptions

bender2k14:


ploeh:
Thank you for writing. I've left [a reply on the page](https://blog.ploeh.dk/2015/05/07/functional-design-is-intrinsically-testable/#904c09a013c144e48a7d322e75f70d94).

PR #590, opened 2020-03-21
added hyperlinks from my comment author lines to my about.me page

bender2k14:
Hello Mark.

Do you mind if I add links from my name on comment author lines to a personal page of mine?

I appreciate your consideration :)


ploeh:
I don't mind at all! It was always my intent that people should leave a link to their personal profiles, but then some stopped doing that, and I realised that not everyone wanted to self-identify like that, so I started to consider it optional.

I do prefer a link, though 👍

bender2k14:
Great! :)

I also prefer to self identify, but I didn't think I had a good enough link to use before. I found about.me recently and very happy with it.

PR #592, opened 2020-03-28
add comment to "The Maybe functor" post

bert2:
> I think it's interesting to note that since C# 8.0 we don't require an extra generic type like `Maybe<T>` anymore in order to implement the maybe functor. Because C# 8.0 added nullable reference types, everything can be nullable now. By adding the right extension methods we can make `T?` a maybe functor and use its beautifully succinct syntax.
>
> Due to the C#'s awkward dichotomy between value and reference types this involves some busy work, so I created a [small nuget package](https://github.com/bert2/Nullable.Extensions) for interested parties.


ploeh:
Thank you for writing 👍 I've left [a response on the page](https://blog.ploeh.dk/2018/03/26/the-maybe-functor/#4f93894d232b485489e9dbba430d510c).

PR #595, opened 2020-04-04
comment replying to Christer about logging events

bender2k14:


bender2k14:
I replied to you @cmeeren

PR #602, opened 2020-05-01
Add response

jfheins:
The PR workflow is a bit more time consuming, but I start to like it for the flexibility ;-)


ploeh:
Thank you for writing. I've left [a response on the page](https://blog.ploeh.dk/2020/04/20/unit-bias-against-collections/#3eed1b53957e4abf8b92c83c15e63ec6).

PR #608, opened 2020-05-28
Question about appropriate dependencies

ledbutter:


ploeh:
Thank you for writing. FWIW, I've left [a reply on the page](https://blog.ploeh.dk/2017/02/02/dependency-rejection/#d0185b75323c4028b04f7dc24e3f127f).

PR #620, opened 2020-06-22
asked if Lazy<T> was intended instead of just T

bender2k14:


ploeh:
Thank you for writing. I've left [a response on the page](https://blog.ploeh.dk/2020/06/22/the-io-functor/#6c7f026e825942429254efc19440b2e2).

PR #621, opened 2020-07-06
comment about difference in two definitions of purity

bender2k14:


ploeh:
Thank you for writing. I've left [a response on the page](https://blog.ploeh.dk/2020/07/06/referential-transparency-of-io/#73bfc62f4f464ac2b5a6131e95884c11).

PR #622, opened 2020-07-06
Comment in discussion about purity definition

bender2k14:


ploeh:
Thank you for writing 👍 I've left [a comment on the page](https://blog.ploeh.dk/2020/02/24/discerning-and-maintaining-purity/#9f9664e29c3f4fafa786abae039eb20e).

PR #623, opened 2020-07-06
comment about the difference between purity and determinism

bender2k14:


ploeh:
Thank you for writing. FWIW, I've left [a response on the page](https://blog.ploeh.dk/2020/07/06/referential-transparency-of-io/#dcd18dda3cf742ffb0bba38ce0326b66).

PR #625, opened 2020-07-09
Referential transparency of IO: You mean discard?

atifaziz:
### Comment

> You can get around the issue using the language's wildcard pattern. This tells the compiler that you're actively ignoring the result of the action.

You made a standard variable declaration and I believe you meant to use a [stand-alone discard] rather than a wildcard pattern? Like below?

  [stand-alone discard]: https://docs.microsoft.com/en-us/dotnet/csharp/discards#a-standalone-discard

```c#
public static string Greet(DateTime now, string name)
{
    _ = Console.WriteLine("Side effect!");
 
    var greeting = "Hello";
    if (IsMorning(now))
        greeting = "Good morning";
    if (IsAfternoon(now))
        greeting = "Good afternoon";
    if (IsEvening(now))
        greeting = "Good evening";
 
    if (string.IsNullOrWhiteSpace(name))
        return $"{greeting}.";
 
    return $"{greeting}, {name.Trim()}.";
}
```

### DIFF
--- a/_posts/2020-07-06-referential-transparency-of-io.html
+++ b/_posts/2020-07-06-referential-transparency-of-io.html
@@ -241,4 +241,40 @@ image_alt: "Diagram showing that the parallel-universe framework executes the IO
 		</div>
 		<div class="comment-date">2020-07-06 19:35 UTC</div>
 	</div>
+
+	<div class="comment" id="247a2c05d83a485a99225057829afa71">
+		<div class="comment-author"><a href="https://www.raboof.com/">Atif Aziz</a></div>
+		<div class="comment-content">
+			<blockquote>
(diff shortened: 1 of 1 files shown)


ploeh:
Thank you for writing. I'll respond later, as I'm a little busy at the moment.

Not a big deal, but the link to https://www.raboof.com/ doesn't work(?)

atifaziz:
> Not a big deal, but the link to https://www.raboof.com/ doesn't work(?)

Ah, it should be `http`, not `https`.

ploeh:
Once again, thank you for writing. I've left [a short comment on the page](https://blog.ploeh.dk/2020/07/06/referential-transparency-of-io/#d0dc32b4548a4d4ab5d55652a7b42502).

PR #627, opened 2020-07-10
comment about determinism vs weak determinism

bender2k14:


ploeh:
Thank you for writing. I've left [a response on the page](https://blog.ploeh.dk/2020/07/06/referential-transparency-of-io/#c53873f100b2480296bfc0ea0709e97d).

PR #629, opened 2020-07-13
comment about use of Lazy vs Func in IO implementation

bender2k14:


ploeh:
Thank you for writing. I've left [a response on the page](https://blog.ploeh.dk/2020/07/13/implementation-of-the-c-io-container/#ba8de284a15c49529bb10390c33d2aa1).

PR #630, opened 2020-07-16
added comment asking why Task<T> is not a functor

bender2k14:


ploeh:
Thank you for writing. I've left [a response on the page](https://blog.ploeh.dk/2018/09/24/asynchronous-functors/#f0bb917b3af24fbe89aac3f4740fbef6).

PR #633, opened 2020-07-23
comment asking why .NET Tasks are not a functor

bender2k14:


ploeh:
Thank you for writing. FWIW, I've left [a response on the page](https://blog.ploeh.dk/2018/09/24/asynchronous-functors/#cffa5d2189d54264b6f6cc58fb84afff).

PR #634, opened 2020-07-25
comment about my previous comment being too critical

bender2k14:


ploeh:
Thank you for writing. I've left [a response on the page](https://blog.ploeh.dk/2020/07/06/referential-transparency-of-io/#dcd28380339f4db8955efd235ef125b6).

PR #638, opened 2020-07-31
Comment for 2018-04-09-coalescing-composite-as-a-monoid.html

arnonax:
Hi Mark,

First I want to explain why I'm writing the comment here: while your blog is fascinating, I find the need for using GitHub pull requests for sending comments very combersome, especially via mobile. I hope you'd understand and edit the comment yourself or direct me to an easier way to post my comment...

And now for the comment itself:
The use of `First<T>` as a mean to make the `Read` method return a monoid is very elegant, but I'm missing something: I don't see a way to extract the actual value out of it... And if there's no value, then you end up in the same situation as before.

Maybe `First<T>` should have another method that takes an `Action<T>` and calls it only if it has a value, but who should call this method and when? 

Can you clarify how one should get the value out of `First<T>`?

Thanks in advance,
Arnon.

### DIFF
--- a/_posts/2018-04-09-coalescing-composite-as-a-monoid.html
+++ b/_posts/2018-04-09-coalescing-composite-as-a-monoid.html
@@ -287,6 +287,7 @@ First {getFirst = Nothing}</pre>
 <div id="comments">
 	<hr>
 	<h2 id="comments-header">Comments</h2>
+<div class="comment">Please see commit message for the actual comment</comment>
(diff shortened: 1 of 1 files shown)


ploeh:
Arnon, thank you for writing. The `First<T>` container is isomorphic to `Maybe<T>` (AKA `Option<T>`). You can read about how to deal with the presence or absence of a value here: https://blog.ploeh.dk/2019/02/04/how-to-get-the-value-out-of-the-monad

HTH

PR #640, opened 2020-08-02
Possibility of pure-only Task<T> functor counterexample

bender2k14:


ploeh:
Thank you for writing. I've left [a response on the page](https://blog.ploeh.dk/2018/09/24/asynchronous-functors/#4f366ad5ff224fbf921c61fd379d26ea).

PR #643, opened 2020-08-03
comment about implementing Tree<>.Max via Aggregate

bender2k14:


ploeh:
Thank you for writing. FWIW, I've left [a response on the page](https://blog.ploeh.dk/2019/06/10/tree-catamorphism/#d832cf82a2984719a0131fa6d802f584).

PR #644, opened 2020-08-03
comment about bug in computing depth of rose tree

bender2k14:


ploeh:
Thank you for writing. FWIW, I've left [a response on the page](https://blog.ploeh.dk/2019/08/05/rose-tree-catamorphism/#bddc949f2130444e89e3d8a73e57935d).

PR #645, opened 2020-08-03
comment about possible fixes to computation of rose tree depth

bender2k14:


ploeh:
Thank you for writing. I've left [a response on the page](https://blog.ploeh.dk/2019/08/05/rose-tree-catamorphism/#1347af54693841828f6ae1e7742311dd).

PR #652, opened 2020-08-23
Are there other patterns beside the Visitor pattern?

wallymathieu:


ploeh:
Thank you for writing. FWIW, I've left [a short reply on the page](https://blog.ploeh.dk/2018/06/25/visitor-as-a-sum-type/#f72ce0f3c59e474b869acded3b542422).

PR #658, opened 2020-09-07
Update 2020-09-07-add-null-check-without-brackets-in-visual-studio.html

james-world:
One-line version of null checking when assigning a field.


ploeh:
Thank you for writing 👍 I've left [a response on the page](https://blog.ploeh.dk/2020/09/07/add-null-check-without-brackets-in-visual-studio/#e456bf65fe3e426f97639b8855622a7b).

PR #660, opened 2020-09-19
Comment on async injection.

votroto:
What a curious way to write comments.

Feels quite uneasy. Hopefully I've understood your intent. Sorry, I couldn't really figure out how to generate a comment id.

### DIFF
--- a/_posts/2019-02-11-asynchronous-injection.html
+++ b/_posts/2019-02-11-asynchronous-injection.html
@@ -992,7 +992,7 @@ public class ReservationsController : ControllerBase
 		<div class="comment-date">2019-06-15 19:25 UTC</div>
 	</div>
 		
-	<div class="comment">
+	<div class="comment" id="9a8a1152bf484fd98804339a4f8239f5">
 		<div class="comment-author">votroto</div>
 		<div class="comment-content">
 			<p>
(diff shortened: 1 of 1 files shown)


TysonMN:
I generate mine [here](https://www.guidgenerator.com/).

votroto:
That's a first... :D. Thanks. 

Does it have to be generated by with the use of a CPRNG, or can I just mash the keyboard until it looks sufficiently random? Can I just copy someone else's? Should I check for uniqueness? Is anyone else going to?

The little README doesn't say. :D

TysonMN:
It is just a [URI fragment](https://en.wikipedia.org/wiki/URI_fragment).  It suffices if it is unique on the page.  You can mash the keyboard if you want.  Then it is easy to check for uniqueness in the page with the "find"  feature of your preferred text editor.

ploeh:
It's fine; I check it before I merge the PR 😄

ploeh:
I'll post a little note here when I've published an answer, but it may be some days...

ploeh:
Thank you for writing. I've left [a response on the page](https://blog.ploeh.dk/2019/02/11/asynchronous-injection/#cb014e7f4c6d420faab1f3328331e7ba).

PR #663, opened 2020-09-30
Add comment to "EnsureSuccessStatusCode as an assertion" post

YevgeniyShunevych:


ploeh:
Thank you for writing. FWIW, I've left [a response on the page](https://blog.ploeh.dk/2020/09/28/ensuresuccessstatuscode-as-an-assertion/#16adadb969f1406c9c18605c8c85ef83).

TysonMN:
I like your response.  Good job Mark.

One small thing. I was supprised by this line.

> ...the information that we care about (the actual status code) is all the way to the right. It'll often require vertical scrolling to see it.

Since you said something was far too the right, I expected you to follow that by saying that it requires _horizontal_ scrolling, not _vertical_ scrolling.  Am I misunderstanding this?  If not, then I think all three uses of "vertical" should be "horizontal".

ploeh:
No, I meant _horizontal_ scrolling; I always get those two mixed up. Feel free to correct my errors 😳

TysonMN:
Glad to :)  See PR #664

YevgeniyShunevych:
Thanks, for the reply, Mark.

I cannot stand aside regarding NUnit. So added another comment PR with lucky winning number #666.

PR #666, opened 2020-10-05
Add another comment to "EnsureSuccessStatusCode as an assertion" post

YevgeniyShunevych:


ploeh:
Thank you for writing; I've left [a short reply on the page](https://blog.ploeh.dk/2020/09/28/ensuresuccessstatuscode-as-an-assertion/#fded96f5f68a4d8ba1e782678dd854be).

PR #670, opened 2020-10-19
comment about nested pairs of parentheses

TysonMN:


ploeh:
Thank you for writing. FWIW, I've left [a short reply on the page](https://blog.ploeh.dk/2020/10/19/monomorphic-functors/#8e6b870ce4aa44aa98a16a7969def6e9).

PR #673, opened 2020-10-26
Fix typo order to orders

TysonMN:
This PR replaces one occurrence of "order" with "orders".  I think "orders" was intended, but maybe I misunderstood something.


ploeh:
Yes, I think that you're right 😄

PR #681, opened 2020-12-07
fixed typo by replacing valid with validity

TysonMN:
I think this is a typo


ploeh:
Thank you for drawing this to my attention. It's, indeed, a typo. I did, however, mean _value_.

PR #686, opened 2020-12-21
added link to Data.Monoid.Endo

TysonMN:
That `Endo` is `Data.Monoid.Endo`, right?  Is this a good link to put there?

### DIFF
--- a/_posts/2020-12-21-a-haskell-proof-of-concept-of-validation-with-partial-data-round-trip.html
+++ b/_posts/2020-12-21-a-haskell-proof-of-concept-of-validation-with-partial-data-round-trip.html
@@ -83,7 +83,7 @@ tags: [Functional Programming, Haskell]
 		The second element of my candidate is <code>[String]</code>, which is fine. Lists are <code>Semigroup</code> instances. But what about <code>Input</code>? Can we somehow combine two <code>Input</code> values into one? It's not entirely clear how we should do that, so that doesn't seem too promising.
 	</p>
 	<p>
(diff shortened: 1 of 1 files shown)


ploeh:
`Endo` is in [Data.Semigroup](https://hackage.haskell.org/package/base/docs/Data-Semigroup.html) in [base](https://hackage.haskell.org/package/base).

When I write Haskell articles, I typically don't link to the types in _base_ or some of the other 'base libraries' such as _time_, _containers_, or _mtl_... I could, but the prose in such an article tends to mention so many of these that I'd have to add a lot of links, and I'm honestly too lazy to add so many.

I'm assuming that the value in doing so is limited, since I think that the majority of the few people who read the Haskell articles know where to find them anyway...

Since you're asking, that assumption might be wrong...

ploeh:
Oh, and if you want to add the link, I always try to link to a version-neutral version of the page, like I've done above...

TysonMN:
I have programed in Haskell just a few times, so many elementary Haskell things are still new to me.  I was curious about specific type class because it is so integral to your prototype.

And thanks for pointing out that I had included version in that URL.  I also prefer to exclude that.

I just force pushed an update.  How is that?

ploeh:
I think that it'd be more appropriate to link to [the definition of Endo](https://hackage.haskell.org/package/base/docs/Data-Semigroup.html#t:Endo).

TysonMN:
The anchor tag in the URL in this PR takes the browser to the line ` Semigroup (Endo a)`, both `Semigorup` and `Endo` have hyperlinks, and the link for `Endo` is the same as the one you have suggested.

TysonMN:
I think you are right though.  `Endo a -> a` is actually a monoid, so the fact that it is a semigroup is immediate.  I will make that change.

TysonMN:
Just force pushed that change.

PR #689, opened 2021-01-06
Specified that a monoid is a set + an operation

arialdomartini:
This PR contains a small change, aimed to provide a slightly more precise definition of a monoid.

In the beginning of your post series you mention that 

> a monoid is a combination of a data type (or set) and an operation.

From there on, though, it seems that you conflate the two components and you only refer to monoids other algebraic structures just as "operations":

> A monoid (not to be confused with a monad) is a binary operation
> Semigroups form a superset of monoids. They are associative binary operations.

That's not quite right. Semigroups are not functions, and they are possibly better described as "algebraic structures consisting of a set together with an associative binary operation".

I think in the narrative this could be a convenient simplification, especially since in FP we always deal with the same category of types.
Sometimes, though, that's a bit misleading. For example, in 

> a monoid is an associative binary operation with a neutral element (also known as identity). 

the reader may wonder how an operation (and not a set) can have a neutral element.

While it's nice only talking of operations in the narrative, I would suggest sticking to more precise descriptions when providing the definition of a concept.

This PR just changes the part of the post where the definition of a monoid is given, repeating that it's actually a set + an operation.

### DIFF
--- a/_posts/2017-10-06-monoids.html
+++ b/_posts/2017-10-06-monoids.html
@@ -150,7 +150,7 @@ image_alt: "Monoids are a subset of semigroups."
 		Summary <a href="#3b10b1afa0d0469392bf89fec90ea371" title="permalink">#</a>
 	</h3>
 	<p>
-		A monoid (not to be confused with a monad) is a binary operation that satisfies the two monoid laws: that the operation is associative, and that an identity element exists. Addition and multiplication are prime examples, but several others exist.
+		A monoid (not to be confused with a monad) is a set (a type) equipped with a binary operation that satisfies the two monoid laws: that the operation is associative, and that an identity element exists. Addition and multiplication are prime examples, but several others exist.
 	</p>
 	<p>
 		(By the way, the identity element for multiplication is <em>one</em> (<em>1</em>), the <em>all</em> monoid is boolean <em>and</em>, and the <em>any</em> monoid is boolean <em>or</em>.)


ploeh:
It's a damned-if-you-do-damned-if-you-don't situation. People come to this topic with varying backgrounds. Some have a mathematical background that enables them to point out details like the ones you've highlighted. Others come with a procedural mindset and are sceptical that mathematics have any relevance to their profession.

I've given workshops based on this material as well, and here attendees also come with varying qualifications. At times, one of the attendees manage to engage me in a detailed discussion similar to the one you imply. If I don't manage that by deferring such a discussion to a break, the eyes of most other attendees glaze over.

There are people (like you) who appreciate the precision, but in my experience, most don't.

I'm aware of mathematical notions of monoids, but I try to keep the language of these articles at least tenuously connected to the reality of 'real' programming. I've tried to keep the language loose enough that it doesn't scare away my average reader. I've mostly used my own words to describe these concepts, since I believe that most of my readers have a background similar to mine. I learned a lot by writing these articles.

From that perspective, I'm not sure that the suggested change is an improvement.

arialdomartini:
> I try to keep the language of these articles at least tenuously connected to the reality of 'real' programming. I've tried to keep the language loose enough that it doesn't scare away my average reader. 

That's why your posts are excellent: the way you explain those topics is oh so amazing. I still have to find a more effective read than yours. I love this post series.

And one of the reasons why they are so effective is exactly because you use a loose-enough language, playing more on intuitiveness and good analogies than on mathematical reasoning.

But I also expect that the very moment you provide a definition ("a monoid is", "a semigroup is"), there, __only there__ you could dare to be just a bit less loose. Odds are readers would read those definitions with a little more attention and expectations anyway, because they are, well, definitions.

ploeh:
Thank you for the kinds words 😄 

Well, it's not like I don't start this particular article like _"Monoids form a subset of semigroups"_ 😅 So my concern that any reader should be scared away by your change, which is _in the conclusion_ is probably unfounded.

PR #690, opened 2021-01-06
Update 2021-01-04-dynamic-test-oracles-for-rho-problems.html

dharmaturtle:
Left a comment with a dramatic ellipses.


ploeh:
Thank you for writing. FWIW, I've left [a short reply on the page](https://blog.ploeh.dk/2021/01/04/dynamic-test-oracles-for-rho-problems/#39b559f49ecc426db7b3d9d1518556df).

PR #696, opened 2021-01-23
Fixed a typo... I think.

dharmaturtle:


ploeh:
Thank you for writing. It's not a typo, though: [Data.Tree](https://hackage.haskell.org/package/containers/docs/Data-Tree.html) _does_ define `Tree a`.

PR #697, opened 2021-01-24
comment about custom map for list not stack overflowing

TysonMN:


ploeh:
Thank you for writing. FWIW I left [a short reply on the page](https://blog.ploeh.dk/2015/12/22/tail-recurse/#ce882ece457a42d0b770f2765a2e21ab).

PR #706, opened 2021-02-20
Update 2018-01-22-function-isomorphisms.html

SpocWeb:
Hi, thank you for your thorough treatise of functional Programming.

I know this is a 3 year old post but I have a Question on Isomorphism between static and OO polymorphic Functions that many others may also have, only they were too shy to ask.

### DIFF
--- a/_posts/2018-01-22-function-isomorphisms.html
+++ b/_posts/2018-01-22-function-isomorphisms.html
@@ -181,3 +181,21 @@ image_alt: "Diagram showing isomorphism between instance method and function."
 		<strong>Next:</strong> <a href="/2018/01/29/argument-list-isomorphisms">Argument list isomorphisms</a>.
 	</p>
 </div>
+	
+<div id="comments">
+	<hr>
+	<h2 id="comments-header">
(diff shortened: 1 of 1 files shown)


ploeh:
Thank you for writing. I've left [a response on the page](https://blog.ploeh.dk/2018/01/22/function-isomorphisms/#13a381feb0924780bcd4ee228616befd).

SpocWeb:
Sorry, if I was not clear. 
What I meant was that there is a Virtual Method Table (I know, Implementation Detail) for each Class Hierarchy and the actual Implementation of a polymorphic is chosen based on the runtime Type of the Object, so if you model Object Methods with Delegates you have to choose the proper Delegate and mimick the Lookup to achieve Polymorphism.

ploeh:
Given that functions/delegates are already polymorphic in C#, F#, and Haskell (the languages that I know best) doesn't the runtime already take care of that?

If you have a higher-order function that takes a `Foo -> Qux -> Corge -> Bar` argument, any function with that signature fits.

PR #707, opened 2021-02-22
A question about explicitly stating "Given" in each test.

belgaard:


ploeh:
Thank you for writing. I've left [a response on the page](https://blog.ploeh.dk/2021/01/25/self-hosted-integration-tests-in-aspnet/#0078721da53e4aa792698fae4a6ab325).

PR #708, opened 2021-02-23
Negative assumptions

belgaard:


ploeh:
Thank you for writing. I've left [a response on the page](https://blog.ploeh.dk/2021/01/25/self-hosted-integration-tests-in-aspnet/#60934ff25b0e45bebca4f00749a6b44c).

PR #709, opened 2021-02-23
Comment about using the new [FromServices] for DI

julealgon:
Adds a comment on `dependency-injection-is-passing-an-argument` in regards to using the `[FromServices]` attribute for injecting dependencies in controllers as method arguments.


ploeh:
Thank you for writing. I've left [a response on the page](https://blog.ploeh.dk/2017/01/27/dependency-injection-is-passing-an-argument/#289838d84c45437ba3d401164f572e37).

PR #710, opened 2021-02-27
Using new Task<T>s that are not started to enforce Method purity

SpocWeb:
Hi, your IO<T> C# Class was very instructive. 
I tried to close the gap you mentioned of Tasks already running and performing side-effects even when their result is ignored 
and found a working SelectMany Implementation that evaluates the Task lazily and only once. 

The idea is to consistently return Tasks that are NOT started and leave it to the Boundary Code to call await and thus trigger all Side Effects. 

I think it would be worth adopting such a practice, at least personally. What do you think?

### DIFF
--- a/_posts/2020-07-27-task-asynchronous-programming-as-an-io-surrogate.html
+++ b/_posts/2020-07-27-task-asynchronous-programming-as-an-io-surrogate.html
@@ -142,3 +142,24 @@ tags: [Functional Programming]
 		The compiler, on the other hand, doesn't help you when it comes to those impure actions that look pure. Neither does it protect you against asynchronous side effects. Diligence, code reviews, and programming discipline are still required if you want to separate pure functions from impure actions.
 	</p>
 </div>
+
+<div id="comments">
+	<hr>
+	<h2 id="comments-header">
+		Comments
+	</h2>
+	<div class="comment" id="c0b2e0bd500a4d5c8407c60bf11bff69">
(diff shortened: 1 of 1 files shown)


SpocWeb:
Maybe you'd like my Test Code to test our this Approach: 


		[Test]
		public static async Task TestTask() {
			var writer = new StringBuilder();
			var timer = GetTime();

			DateTime Selector_(DateTime t) {
				_ = writer.Append(t).AppendLine();
				return t; //new IO<DateTime>(() => t);
			}

			Task<DateTime> io2 = timer.SelectMany(x
				=> new Task<DateTime>(() =>Selector_(x)));
			var length = writer.Length.ShouldBe(0);
			_ = await io2; //This triggers all Queries and Side-Effects at once!
			length = writer.Length.ShouldBeMoreThan(length);
			_ = await io2; //This triggers all Queries and Side-Effects at once!
			_ = writer.Length.ShouldBe(length);
			io2 = timer.SelectMany(x
				=> new Task<DateTime>(() =>Selector_(x)));
			_ = writer.Length.ShouldBe(length);
			_ = await io2; //This triggers all Queries and Side-Effects at once!
			length = writer.Length.ShouldBeMoreThan(length);
		}

ploeh:
Thank you for writing. FWIW I've left [a short reply on the page](https://blog.ploeh.dk/2020/07/27/task-asynchronous-programming-as-an-io-surrogate/#35f7c2e8cdc84a89b79771f541530025).

PR #711, opened 2021-02-28
Details in arrange

belgaard:


ploeh:
Thank you for writing. I've left [a response on the page](https://blog.ploeh.dk/2021/01/25/self-hosted-integration-tests-in-aspnet/#76bca252379c4f73a676ea505f16a550).

PR #712, opened 2021-03-13
Possible Restoration of Functor Law for HashSets

SpocWeb:
Hi again, I'm still plowing though your excellent Functional Series and thought of some Resolutions to the broken functor Laws of HashSet. 
I think the Difference between weak Equality and strong Identity ist important here.


ploeh:
Thank you for writing. I've left [a short response on the page](https://blog.ploeh.dk/2018/12/03/set-is-not-a-functor/#2e5d7f35754e4b1b9ae2ffc2ab353be1).

PR #713, opened 2021-03-16
HashSet is a functor

TysonMN:


ploeh:
Thank you for writing. I've left [a response on the page](https://blog.ploeh.dk/2018/12/03/set-is-not-a-functor/#11b5047c500241058aa9a58320157824).

PR #714, opened 2021-03-16
Answer to DateTimeOffset.EqualsExact 

SpocWeb:
Yes, of course we can use DateTimeOffset.EqualsExact
But it is only one line and we have to re-implement GetHashCode anyway :-)


ploeh:
Thank you for writing. It's not your fault, but as it sometimes happens, your commit arrived concurrently with another, so there's a merge conflict. Could you please address it?

TysonMN:
> Ah, so you're proposing [...]

@SpocWeb, to whom is "you"referring?

TysonMN:
Also, the comment IDs are the same.

PR #722, opened 2021-04-12
fixed typo by replacing FileInfo with string

TysonMN:


ploeh:
Good catch 👍  Thank you

PR #728, opened 2021-05-01
Comment on Leaky abstraction by omission

lufthavn:


ploeh:
Thank you for writing. I've left [a response on the page](https://blog.ploeh.dk/2021/04/26/leaky-abstraction-by-omission/#c30f63e0ba234d15acecb40800efc0ec).

PR #729, opened 2021-05-02
Comment on Leaky abstraction by omission

thibautde:


ploeh:
Thank you for writing. I've left [a response on the page](https://blog.ploeh.dk/2021/04/26/leaky-abstraction-by-omission/#3104d83e33524adc9f4d349fa4df20ed).

PR #734, opened 2021-05-27
Update 2021-05-24-tennis-kata-using-the-state-pattern.html

JimCooperFromGmailAccount:
A state machine seem too complex for a simple problem

### DIFF
--- a/_posts/2021-05-24-tennis-kata-using-the-state-pattern.html
+++ b/_posts/2021-05-24-tennis-kata-using-the-state-pattern.html
@@ -296,3 +296,117 @@ tags: [Unit Testing, Design Patterns]
 		Once you recognise that a problem can be modelled as a finite state machine, you have new options. If you're in a conservative context where colleagues aren't keen on fancy FP shenanigans, you can always reach for the State design pattern.
 	</p>
 </div>
+
+
+
+Do you think that perhaps you are at risk of making too many problems look like nails for your state machine hammer? :-)
+Actually, you just want to convert a pair of points into a tennis score. That doesn't require a state machine, I don't think:
+
+<pre>
+
+using NUnit.Framework;
+
+namespace TennisKata
+{
+    public class Tests
+    {
+        private TennisGame tennisGame;
+
+        [SetUp]
+        public void Setup()
+        {
+            tennisGame = new TennisGame();
+        }
+
+        [TestCase(0, 0, ExpectedResult = "Love All")]
+        [TestCase(1, 1, ExpectedResult = "Fifteen All")]
+        [TestCase(2, 2, ExpectedResult = "Thirty All")]
+        [TestCase(3, 3, ExpectedResult = "Deuce")]
+        [TestCase(4, 4, ExpectedResult = "Deuce")]
+        [TestCase(1, 0, ExpectedResult = "Fifteen - Love")]
+        [TestCase(2, 1, ExpectedResult = "Thirty - Fifteen")]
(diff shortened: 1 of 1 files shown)


ploeh:
Thank you for writing. I've left [a response on the page](https://blog.ploeh.dk/2021/05/24/tennis-kata-using-the-state-pattern/#37f19a75fa454aae91a67b4b2a573ada).

JimCooperFromGmailAccount:
Hi Mark

Thanks for the reply. I would point out that translation is as trivial/complex in my solution as in any other. It’s a pretty orthogonal issue. For the players names, you just need the server and receiver names (I left that out to simplify the code, actually). But the only method that needs them is already parameterised, even :-)

I do understand the point of trying out things, but there is sentence in your article that caught my eye:

"After having spent the majority of the last decade with functional programming, I've come to realise that many problems are really just state machines waiting to be revealed as such”

It immediately made me think "uh-oh, this sounds like a hammer”. We’ve both seen many example of favourite techniques being over/inappropriately used (regexes spring to mind). Having been guilty of it myself, I would not like to encourage it any further :-)

And, secondly, it does not seem to me to actually model the problem. I don’t believe a game of tennis has many states in reality. Only 2 really: not decided, and won (abandoned would be a state for the entire match). The only thing important during a game is the score. It’s trivially easy to base an API on my sort of code then (not using the score string, obviously - why would you?) :-)

I guess you would want to try and put it into context e.g in a scoreboard system for a match (including sets, players names, doubles/singles, when to change ends and balls maybe, and so on) to evaluate whatever technique you used. Also, in some tournaments, there are tiebreak games that have a different scoring method. That’s maybe a step or two beyond a simple code kata*, although maybe it’d be a useful exercise sometimes. I would then pose the question, if this was a general purpose scoreboard for multiple sports, even similar ones like badminton and squash, what would you do differently then?

I see little value in coming up with more complex solutions than necessary. I have had plenty of colleagues do that for me, over the decades. I have not generally appreciated their efforts :-) 

I was very strict with TDD, and that’s what I came up with. A state machine (or FSA) is pretty over the top, IMO, and I can’t see the path where I personally would have ever got to one. However, I plan to run a pair/mob programming exercise with some of my current colleagues, and I might use this kata. I’ll be interested to see how complicated they make things :-)

In a weird sort of coincidence, I’m currently working on/struggling with some of their code where they probably should have used the State pattern, but haven’t…

Cheers
Jim Cooper

* I dislike this name, personally. It’s based on a wide-spread misunderstanding of what a kata is. 

** Similarly, I’ve always thought the people who named Scrum had never seen a game of rugby :-) Eight big hairy blokes grabbing each other by the balls in the rain and mud, heaving against eight other big hairy blokes, and getting nowhere before collapsing in a huge confused heap and getting covered in crap, does seem an accurate metaphor for some software development processes I’ve seen, though!

ploeh:
Oh, there are plenty of bad metaphors around. Extending your comment on Scrum, calling each sustained effort a *sprint* also seems quite misguided.

It's not my intent to apply a Golden Hammer. I rarely reach for the State pattern or its functional equivalents as the first thing. If there's a simpler solution, I go for that.

Doing katas (yes, I'm aware that the name is another bad metaphor) is deliberate practice. The intent is to become aware of a solution before one needs it. As Milton Friedman said in a completely different context:

> "Only a crisis – actual or perceived – produces real change. When that crisis occurs, the actions that are taken depend on the ideas that are lying around"

In this light, I view badly factored code as a crisis. When you hit a point where you have to do something about legacy code, you reach for the ideas that are lying around - i.e. the techniques you know.

PR #738, opened 2021-06-21
Update 2021-06-21-agile-pull-requests.html

Hettomei:
We need to move this engineering work into the scientific approach. Maybe.


ploeh:
Thank you for writing. I've left [a response on the page](https://blog.ploeh.dk/2021/06/21/agile-pull-requests/#89eeaad6eded422b859e901e1157a589).

PR #739, opened 2021-06-23
Comment: book availability in europe

ericvruder:
A new comment about when the book would be generally available in Europe.


ploeh:
Thank you for writing. I've left [a response on the page](https://blog.ploeh.dk/2021/06/14/new-book-code-that-fits-in-your-head/#7cfe04d6068f49a88db5e7317f0fe11d).

PR #742, opened 2021-06-28
Comment about eBook availability

srogovtsev:


ploeh:
Thank you for writing. I had to ask the publisher, but I've now added [a short response to the page](https://blog.ploeh.dk/2021/06/14/new-book-code-that-fits-in-your-head/#6ef395bd0b3b48299861798582d2e492).

srogovtsev:
If you don't mind me asking, who'll be distributing the PDF version? Kindle I can get from, well, Amazon Kindle, but PDFs are trickier, and one should usually go to the actual publisher for that.

(I also hope to be able to get access to your book in O'Reilly library, but that will depend on their agreement with your publisher as well)

ploeh:
I don't know the answers to these questions. The publisher handles production of both print and ebook. Given that this is part of [the Robert C. Martin series](https://www.informit.com/imprint/series_detail.aspx?ser=348084), it'd be natural if they handle these things like the other books in the series, but that's just my guess.

srogovtsev:
Thank you for the information!

PR #750, opened 2021-08-09
Fix typos

bartelink:


ploeh:
Thank you. You can tell that this was a difficult post to write by the more-frequent-than-usual occurrences of typos 😀

PR #755, opened 2021-08-24
Fix broken links pointing to MSDN blog

farlee2121:
Microsoft must have moved domains from blogs.msdn.com to docs.microsoft.com. The link structure also changed and specific post links didn't forward correctly.


ploeh:
Good detective work tracking down those new addresses. Thank you.

PR #756, opened 2021-09-14
monomorphic functors should be called invariant functors

TysonMN:


ploeh:
Thank you for writing. I've left [a response on the page](https://blog.ploeh.dk/2020/10/19/monomorphic-functors/#3462f8e2c4a44593889f4816ba41ed88).

PR #759, opened 2021-09-20
Add comment

cmeeren:
Side note: I still find the comment system rather tedious. I often have something I want to say and I think would contribute to the discussion, but having to find an old post with comments, copy the "template" into the new post, and replace the contents with a comment I write in plain HTML is enough of a strain on my resources that I mostly choose not to do it.

I have no illusions that this will be changed; I just wanted to get it out there. 🙂


ploeh:
Thank you for writing. I've left [a response on the page](https://blog.ploeh.dk/2021/09/20/keep-ids-internal-with-rest/#2fa9b78a83b64b33ae16182ed7db450c).

PR #762, opened 2021-10-04
Fix typo in 2021-10-04-reader-as-a-contravariant-functor

DanCarterGMSL:
I think I've found a mistake in this post, and I believe this PR fixes it.


ploeh:
Yes, indeed, you're right! Good catch. Thank you.

That's what happens when you become a little too copy-and-paste-happy. I even have unit tests of both methods, but alas, the one covering one method was a copy-and-paste of the other test 🤦‍♂️

PR #766, opened 2021-11-15
Add comment about infinite sets with different cardinalities

TysonMN:


ploeh:
Thank you for writing. I've left [a comment on the page](https://blog.ploeh.dk/2021/11/15/types-as-sets/#ff1df88a06aa42349fb56e243299499c).

PR #769, opened 2021-12-10
LSP comment about subtype polymorphism

TysonMN:


ploeh:
Thank you for writing. I've left [a response on the page](https://blog.ploeh.dk/2021/12/06/the-liskov-substitution-principle-as-a-profunctor/#c4279163948e4fda98899fa54a6cfc16).

PR #772, opened 2021-12-17
The readRangeBySql query looks wrong to me

m-ringler:
The date comparison here does not seem to be correct, what you want is 'where MinDate <= Date && Date <= MaxDate',
 which would be true for 2021-12-31 / 2022-01-01 / 2022-01-01, but your test is not


ploeh:
You're right. Thank you for pointing that out. I no longer recall why I wrote the date comparison like that back in 2018, but FWIW in the codebase that supports my book [Code That Fits in Your Head](https://blog.ploeh.dk/code-that-fits-in-your-head), it looks like this:

```csharp
private const string readByRangeSql = @"
    SELECT [PublicId], [At], [Name], [Email], [Quantity]
    FROM [dbo].[Reservations]
    WHERE [RestaurantId] = @RestaurantId AND
            @Min <= [At] AND [At] <= @Max";
```

PR #777, opened 2022-01-13
💬 Add comment regarding readability benefits to argument validation simplification

julealgon:


ploeh:
Thank you for writing. I've left [a response on the page](https://blog.ploeh.dk/2014/08/07/why-dry/#d6608e95d8d349fe802d5e0ba02f0c31).

PR #784, opened 2022-02-03
Local function to enforce balance

Aszan:
I don't find a proper way to format my snippet. How do you proceed?
Is there a way to embed a gist?


ploeh:
If you don't have a code-to-HTML formatter handy, you can always just surround the code with a `<pre>` tag... I've done it for you, though 😄 

I'll write a response later...

ploeh:
Thank you for writing. I've left [a response on the page](https://blog.ploeh.dk/2013/06/24/a-heuristic-for-formatting-code-according-to-the-aaa-pattern/#c9a14538f4154f71a5ae17a50cb5f753).

PR #787, opened 2022-02-22
Test reuse comment

farlee2121:


ploeh:
Thank you for writing. I've left [a response on the page](https://blog.ploeh.dk/2014/10/09/unit-testing-methods-with-the-same-implemenation/#c9f45b24cb5944e5934acd86921f4511).

PR #791, opened 2022-03-01
Add comment on flattening statically typed lists

michens:
Is there a way to generate or preview these comments? I tried following examples, but I have no idea if this is right


ploeh:
Unless you want to install Jekyll, I'm afraid there's no preview feature.

Your comment looks good, though 👍

PR #799, opened 2022-04-12
Comment on 2022-04-04-kleisli-composition.html

GiuseppePiscopo:


ploeh:
Thank you, the comment is fine, and I'll be happy to publish it if you'd like.

Usually, people just send me a pull request with a fix when they spot a typo. It's not normally something people use a comment for, but as I wrote, I'm happy to post it if you'd like me to.

In any case, thank you for pointing out the typo. I've already corrected it.

GiuseppePiscopo:
d'oh!

That's right. Given that I was submitting a PR, I could as well have proposed the fix straight into the actual page content. It did not came to my mind.

Well, I guess the comment is not needed by now. I'll close this PR then.

Thanks for the greate contents!

PR #800, opened 2022-04-18
Add comment about generic vs non-generic functors

TysonMN:


ploeh:
True, the illustration is a simplification.

TysonMN:
I don't think it is worth changing the image though since this detail is not relevant to the topic of monads.

PR #801, opened 2022-04-18
Add comment about relationship between monad and monoid

TysonMN:


ploeh:
Thank you for writing. I've left [a response on the page](https://blog.ploeh.dk/2022/04/11/monad-laws/#2bf9e517fa744a89928ae3d9f08fedd2).

PR #810, opened 2022-05-17
Division semantics fixes

mormegil-cz:
I think this was the intended (expected) meaning: 1 / int is integer division in C#, so 1/2 == 0. (Although the current version does not prevent the point of the article: Even the piecewise-constant functions fulfill the monad laws…)


ploeh:
Yes, you're right, good catch! Thank you 👍

PR #814, opened 2022-05-26
Explain background of spec-like testing

farlee2121:
Deeper explanation of random testing with constrained static types.

The code blocks are a bit awkward. I can find a way to colorize them and eliminate extra spacing if you'd like.


ploeh:
I mostly consider comments 'islands of user content', so as long as you don't break how the page renders, or include content of illegal nature, etc. (to be decided at my discretion), you have a great degree of freedom. Let the code remain as is or format it as you see fit 😄 It's readable enough already.

ploeh:
Thank you for writing. I've left [a response on the page](https://blog.ploeh.dk/2015/05/07/functional-design-is-intrinsically-testable/#5c4d67c21b7048f2a423dc67f2d04caa).

PR #816, opened 2022-05-27
Respond to spec property thread

farlee2121:


ploeh:
Thank you for writing. I've left [a response on the page](https://blog.ploeh.dk/2015/05/07/functional-design-is-intrinsically-testable/#9209b510035645399e2e720228af9271).

PR #835, opened 2022-09-17
Add comment about structural typing

TysonMN:


ploeh:
Thank you for writing. I've left [a response on the page](https://blog.ploeh.dk/2022/08/01/invariant-functors/#98be00b677484b11a51d8e583c25baa5).

PR #836, opened 2022-09-17
Add comment about applicative behavior for a tree

TysonMN:


ploeh:
Thank you for writing. I've left [a short response on the page](https://blog.ploeh.dk/2019/06/10/tree-catamorphism/#0b6120ae739747c3b040bb5734102ef8).

PR #838, opened 2022-09-17
Add comment asking if Post method is an impureim sandwich

TysonMN:


ploeh:
Thank you for writing. I've left [a response on the page](https://blog.ploeh.dk/2022/09/12/coalescing-dtos/#2352255931b1499184b5ba3782436d61).

PR #841, opened 2022-09-24
Add comment challenging red-green-refactor.

danielt1263:


ploeh:
Thank you for writing. I've left [a comment on the page](https://blog.ploeh.dk/2022/09/19/when-to-refactor/#f9df3a46ce5f4cb8a386304f64030137).

TysonMN:
@danielt1263, I also added [a response](https://github.com/ploeh/ploeh.github.com/pull/846).

danielt1263:
Hmm... Just like in language composition class, the first draft of our code is never the one that we submit. I don't think of that as refactoring, but I'm willing to concede the point if you want to call that refactoring.

What I'm against, is this idea of going into a code base just to "refactor". Maybe the current code isn't at the right level of abstraction or maybe it is, but without a concrete feature to add any change in abstraction level is just guessing.

TysonMN:
Writing the first draft is not refactoring. It is the "red" and "green" steps of "red, green, refactor".  The refactor happens when you already know where you are going and you want to bring other people along with you.

In this way, it is much like teaching. You shouldn't spend time fixing spelling mistakes in your lecture notes until you have sufficiently understood the material.

> What I'm against, is this idea of going into a code base just to "refactor".

I agree that this should not usually happen.

Do you encounter people that claim it is often a good idea just to factor?

danielt1263:
I constantly encounter people who complain that they want to refactor the code base but their manager won't give them the time to do it. They aren't saying they want to add a new feature but the manager won't give them time. They aren't saying they want to fix a bug but don't have time. They aren't saying they don't have time to make a final draft of the code they are currently working on. They are saying they don't have time to "refactor". 

What I submit is that many times these people are wanting to change the code in order to add some level of abstraction. They want to "make future changes as easy as possible" (as Mark said) but they don't actually have any particular "future change" in mind. This sort of refactoring is dangerous. It's change merely for the sake of change and doesn't IMO improve the code; instead it tends to obfuscate it.

TysonMN:
I completely agree

PR #849, opened 2022-10-27
Encapsulation: comment about demonstrating whole idea in one language

atifaziz:
This PR adds the following comment to [Encapsulation in Functional Programming](https://blog.ploeh.dk/2022/10/24/encapsulation-in-functional-programming/):

> I wonder what's the goal of illustrating OOP-ish examples exclusively in C# and FP-ish ones in F# when you could stick to just one language for the reader. It might not always be as effective depending on the topic, but for encapsulation and the examples shown in this article, a C# version would read just as effective as an F# one. I mean when you get round to making your points in the **Immutable Table** section of your article, you could demonstrate the ideas with a C# version that's nearly identical to and reads as succinct as the F# version:
>
> ```c#
> #nullable enable
>
> readonly record struct Reservation(int Quantity);
> abstract record Table;
> record StandardTable(int Capacity, Reservation? Reservation): Table;
> record CommunalTable(int Capacity, ImmutableArray<Reservation> Reservations): Table;
> 
> static class TableModule
> {
>     public static StandardTable? Standard(int capacity) =>
>         0 < capacity ? new StandardTable(capacity, null) : null;
> 
>     public static CommunalTable? Communal(int capacity) =>
>         0 < capacity ? new CommunalTable(capacity, ImmutableArray<Reservation>.Empty) : null;
> 
>     public static int RemainingSeats(this Table table) => table switch
>     {
>         StandardTable { Reservation: null } t => t.Capacity,
>         StandardTable => 0,
>         CommunalTable t => t.Capacity - t.Reservations.Sum(r => r.Quantity)
>     };
> 
>     public static Table? Reserve(this Table table, Reservation r) => table switch
>     {
>         StandardTable t when r.Quantity <= t.RemainingSeats() => t with { Reservation = r },
>         CommunalTable t when r.Quantity <= t.RemainingSeats() => t with { Reservations = t.Reservations.Add(r) },
>         _ => null,
>     };
> }
> ```
>
> This way, I can just point someone to your article for enlightenment, &#x1F609; but not leave them feeling frustrated that they need F# to (practice and) model around data instead of state mutating objects. It might still be worthwhile to show an F# version to draw the similarities and also call out some differences; like `Table` being a true discriminated union in F#, and while it appears to be emulated in C#, they desugar to the same thing in terms of CLR types and hierarchies.
>
> By the way, in the C# example above, I modeled the standard table variant differently because if it can hold only one reservation at a time then the model should reflect that.

### DIFF
--- a/_posts/2022-10-24-encapsulation-in-functional-programming.html
+++ b/_posts/2022-10-24-encapsulation-in-functional-programming.html
@@ -230,3 +230,48 @@ tags: [Functional Programming, F#]
 		Perhaps you can find a functional programmer who might be slightly offended if you suggest that he or she should consider encapsulation. If so, suggest instead that he or she considers the properties of functions and data.
 	</p>
 </div>
+<div class="comment" id="bb616acbc1ac41cb8f937fe7175ce061">
+	<div class="comment-author"><a href="http://www.raboof.com/">Atif Aziz</a></div>
+	<div class="comment-content">
+		<p>
+			I wonder what's the goal of illustrating OOP-ish examples exclusively in C# and FP-ish ones in F# when you could stick to just one language for the reader. It might not always be as effective depending on the topic, but for encapsulation and the examples shown in this article, a C# version would read just as effective as an F# one. I mean when you get round to making your points in the <strong>Immutable Table</strong> section of your article, you could demonstrate the ideas with a C# version that's nearly identical to and reads as succinct as the F# version:
+		</p>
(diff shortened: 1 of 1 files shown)


ploeh:
Thank you for writing. I've left [a response on the page](https://blog.ploeh.dk/2022/10/24/encapsulation-in-functional-programming/#7069ea2b33a64a1caf7247c3a1543bac).

PR #851, opened 2022-11-02
question on coverage while using FakeDatabase added

Aarskin:


ploeh:
Thank you for writing. I've left [a response on the page](https://blog.ploeh.dk/2022/10/17/stubs-and-mocks-break-encapsulation/#98ef29aec08e408382205b49bea697b6). HTH

PR #857, opened 2023-01-02
Change appdata to AppData

moodmosaic:
That is how it's also written in the docs, as it seems: https://learn.microsoft.com/en-us/windows/win32/shell/csidl


moodmosaic:
Neat article ✈️

ploeh:
That is indeed what the directory is called on disk...

PR #860, opened 2023-01-16
Add comment about applicative assertions

TysonMN:


ploeh:
Thank you for writing. I've left [a response on the page](https://blog.ploeh.dk/2022/11/07/applicative-assertions/#a1516515c1b84cb2b8473f2c0321562e).

PR #871, opened 2023-02-27
comment on Test Data Generator monad

AnthonyLloyd:


ploeh:
Thank you for writing. I've left [a short response on the page](https://blog.ploeh.dk/2023/02/27/test-data-generator-monad/#21f1a20a1a5944fbbbdd300c8f0e89b0).

AnthonyLloyd:
Thank you. I enjoyed you post on applicative assertions recently. I must admit F# is the best language for testing. CEs give mutliple features like applicative assertions, logging, async in a much better/simpler/faster way. I am only able to use it though in specific places I can justify (like MKL.NET). Other developers tend to be a bit more TestAdapter focused which is a limitation.

PR #878, opened 2023-04-03
Comment on An abstract example of refactoring from interaction-based to property-based testing

srogovtsev:


ploeh:
Thank you for writing. I've left [a response on the page](https://blog.ploeh.dk/2023/04/03/an-abstract-example-of-refactoring-from-interaction-based-to-property-based-testing/#076f05eef3f649eab07392dc2ac83023).

PR #884, opened 2023-04-10
Comment on An abstract example of refactoring from interaction-based …

srogovtsev:
…to property-based testing


ploeh:
Thank you for writing. I've left [a response on the page](https://blog.ploeh.dk/2023/04/03/an-abstract-example-of-refactoring-from-interaction-based-to-property-based-testing/#ee233e2d67954197afcaa98d7637fda1).

PR #891, opened 2023-04-23
Comment on 2013-10-23-mocks-for-commands-stubs-for-queries.html

marcinjahn:


ploeh:
Thank you for writing. I've left [a response on the page](https://blog.ploeh.dk/2013/10/23/mocks-for-commands-stubs-for-queries/#3e74a22c05db4bbe83f863d9d017ff0b).

PR #907, opened 2023-08-07
2013/10/23/mocks-for-commands-stubs-for-queries Possible fix for example code

VadymLevkovskyi:
Hello, @ploeh 

I'm, not sure if it was intentional from your side, but it looks like there is minor issue in code in https://blog.ploeh.dk/2013/10/23/mocks-for-commands-stubs-for-queries/ in next section

< cut from the mentioned article >

# Data Flow Verification 

... (skipped)

With the Stub verifying the Query (GetUserReturnsCorrectValue), and the other test that uses a Mock to verify a Command (UserIsSavedIfItDoesNotExist), the final implementation is:

```
public User GetUser(int userId)
{
    var u = this.userRepository.Read(userId);
    if (u.Id == 0)
        this.userRepository.Create(1234);
    return u;
}
```
< cut from the mentioned article >

where `1234` is used as `user id` in `userRepository.Create()`. 

Also, it's interesting, that this isssue is not caught by unit test mentioned in the same section, that makes me worry about the whole story. And yes, I'm aware that `this.userRepository.Create(1234);` is a `command`, not a `query`, and thus it has to be tested by `mock`, not a `stub`. If possible, I'd like this fact to be (at least) mentioned separately. Or (minimal efforts principle) "fixed" (hiding the problem, in fact) by changes proposed here.

Sorry for necroposting, hope it was valuable.

### DIFF
--- a/_posts/2013-10-23-mocks-for-commands-stubs-for-queries.html
+++ b/_posts/2013-10-23-mocks-for-commands-stubs-for-queries.html
@@ -310,7 +310,7 @@ tags: [Unit Testing]
 {
 &nbsp;&nbsp;&nbsp; <span style="color: blue;">var</span> u = <span style="color: blue;">this</span>.userRepository.Read(userId);
 &nbsp;&nbsp;&nbsp; <span style="color: blue;">if</span> (u.Id == 0)
-&nbsp;&nbsp;&nbsp; &nbsp;&nbsp;&nbsp; <span style="color: blue;">this</span>.userRepository.Create(1234);
+&nbsp;&nbsp;&nbsp; &nbsp;&nbsp;&nbsp; <span style="color: blue;">this</span>.userRepository.Create(userId);
 &nbsp;&nbsp;&nbsp; <span style="color: blue;">return</span> u;
 }</pre>
 	</p>


ploeh:
Thank you for writing.

I admit that this could have been clearer. I had to consult the source code I actually used for that article, which is from 2013. In it, there's another test that looks like this:

```csharp
[Fact]
public void UserIsSavedIfItDoesNotExist()
{
    var td = new Mock<IUserRepository>();
    td.Setup(r => r.Read(1234)).Returns(new User { Id = 0 });
    var sut = new SomeController(td.Object);

    sut.GetUser(1234);

    td.Verify(r => r.Create(1234));
}
```

While it does test that `Create` was called, it only verifies that it was called with the constant value `1234`. That's the reason that all tests pass.

The implementation code is still incorrect, which indicates that there's at least one missing test case. The immediate remedy, then, is to add another test case which would then elicit a fix of the issue that you've called out.

FWIW, I no longer rely on Mocks and Stubs for unit testing, so while I still consider this article correct, I don't think the suggested solution is optimal. My 2019 article [From interaction-based to state-based testing](https://blog.ploeh.dk/2019/02/18/from-interaction-based-to-state-based-testing) better captures my current approach.

VadymLevkovskyi:
Thanks man, checking that new approach right now! 

And I see, that you had mentioned it in the comment under the original article, so this PR is no longer relevant, thus closing it.

ploeh:
I hadn't really looked at that article for years, but it seems a follow-up is warranted. I plan to write another article to that effect.

ploeh:
Follow-up article published: https://blog.ploeh.dk/2023/08/14/replacing-mock-and-stub-with-a-fake

Thank you for the inspiration 🙏

PR #908, opened 2023-08-14
fix: still creating a user with wrong id in the fake

jlauwers:
Hello, thanks for your post 😀

In the fake repo, you still create a user with a wrong id.
It should be `return new User { Id = userId  };`

```
public sealed class FakeUserRepository : Collection<User>, IUserRepository
{
    public void Create(int userId)
    {
        Add(new User { Id = userId });
    }
 
    public User Read(int userId)
    {
        var user = this.SingleOrDefault(u => u.Id == userId);
        if (user == null)
            return new User { Id = 0 };
        return user;
    }
}
```


jlauwers:
Nevermind... 
It's "A typical Fake" exemple... so not keeping the creation of a "real" new user.

PR #912, opened 2023-09-21
Add open graph meta tags

chamook:
Pages already have the twitter specific tags that enhance the way the link preview is rendered if a link is posted there, but they are not supported by all platforms. This adds the corresponding open graph tags that serve the same function on other sites (including Mastodon which is where I noticed this discrepancy)


ploeh:
Thank you, I hadn't really thought about this, but it seems like a really good idea 👍

PR #918, opened 2023-10-09
Update 2023-09-18-do-orms-reduce-the-need-for-mapping.html

opcodewriter:
My comment


ploeh:
Thank you for writing. I left [a response on the page](https://blog.ploeh.dk/2023/09/18/do-orms-reduce-the-need-for-mapping/#0c76c456b47e42ec872603996ba1cfc0).

PR #920, opened 2023-10-11
Add comment to 2023-10-02-dependency-whac-a-mole.html

oakular:
Adding some reflections on the post.


ploeh:
Thank you for writing. I've added [a response to the page](https://blog.ploeh.dk/2023/10/02/dependency-whac-a-mole/#eef76159a60b4ee482238b1cd990ab94).

PR #925, opened 2023-10-29
comment on 2018-04-23-null-object-as-identity

catwang01:
Leave comment on the post 2018-04-23-null-object-as-identity


ploeh:
Thank you for writing. I've left [a response on the page](https://blog.ploeh.dk/2018/04/23/null-object-as-identity/#c72b8d0217164104a1a962b6476dea11).

PR #928, opened 2023-12-05
Precise:  Serialization w/ & without Reflection

blackheaven:
Haskell has reflection through [Typeable](https://hackage.haskell.org/package/base-4.19.0.0/docs/Data-Typeable.html#t:Typeable), however it's not widely use.

I tend to use it at the boundaries of my Domain, when I have runtime types uncertainty.

I agree that they are not used for serialization (as far as I know), and that the RTS does not have enough knowledge on the types it deals with (and we cannot access to what it knows).

### DIFF
--- a/_posts/2023-12-04-serialization-with-and-without-reflection.html
+++ b/_posts/2023-12-04-serialization-with-and-without-reflection.html
@@ -156,3 +156,44 @@ image_alt: "Bar seating at Ernst, Berlin."
         <strong>Next:</strong> Serializing restaurant tables in Haskell.
     </p>
 </div>
+
+<div id="comments">
+    <hr>
+    <h2 id="comments-header">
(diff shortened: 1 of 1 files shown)


ploeh:
I stand corrected.

Could you change that correction to a comment on the page?

I distinguish between who asserts what, and even when I'm mistaken (as it seems that I am in this case), I find it more honest that readers understand the source of knowledge (or lack thereof) in the articles.

blackheaven:
Thanks, here's a new draft, let me know if I have things to change.

ploeh:
Thank you for writing. I've left [a reply on the page](https://blog.ploeh.dk/2023/12/04/serialization-with-and-without-reflection/#9cf7ca25c1364eb39c489d9883d90cea).

PR #929, opened 2023-12-11
Add-comment: Serialization w/ & without Reflection (#928 - follow-up)

blackheaven:
Sorry for the delay.

How do you generate the syntax highlighting?

### DIFF
--- a/_posts/2023-12-04-serialization-with-and-without-reflection.html
+++ b/_posts/2023-12-04-serialization-with-and-without-reflection.html
@@ -209,4 +209,36 @@ image_alt: "Bar seating at Ernst, Berlin."
         </div>
         <div class="comment-date">2023-12-08 17:11 UTC</div>
     </div>
+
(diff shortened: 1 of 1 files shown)


ploeh:
> Sorry for the delay.

No need to apologize. The discussion in the comments happens at the pace with which we can manage 😄 

> How do you generate the syntax highlighting?

To be clear, I consider that entirely optional for comments, but I use [Copy as HTML](https://marketplace.visualstudio.com/items?itemName=VisualStudioPlatformTeam.CopyAsHtml) - a Visual Studio (not Code) add-in. And, believe it or not, with another add-in, [Haskell Tools](https://marketplace.visualstudio.com/items?itemName=KristianSkovJohansen.VSHaskellTools), I can get syntax highlighting of Haskell code.

I don't develop Haskell in Visual Studio (I use VS Code for that), but to preserve the HTML syntax colouring, I haven't found a better combo. Not that I haven't tried.

PR #931, opened 2024-01-09
Add comment to 2024-01-08-a-range-kata-implementation-in-haskell

mormegil-cz:


ploeh:
Thank you for writing. I've added [a response on the page](https://blog.ploeh.dk/2024/01/08/a-range-kata-implementation-in-haskell/#9d0f60b0a2654424b10d264cfd8b6c96).

PR #932, opened 2024-01-16
Add a comment for the replacing-mock-and-stub-with-a-fake article

AmiradelBeyg:
Hi Mark,

I previously posted my question regarding Fakes and test exceptions in X on Twitter. As you suggested, I've now created a pull request on the blog, albeit with a slight delay. I hope this aligns with the correct structure you were expecting. Additionally, I've included more details regarding the question.

Please inform me if further changes or additional details are required.

Thanks,
Amir

### DIFF
--- a/_posts/2010-02-03-ServiceLocatorisanAnti-Pattern.html
+++ b/_posts/2010-02-03-ServiceLocatorisanAnti-Pattern.html
@@ -1246,35 +1246,3 @@ The fundamental thing to be aware of, however, is that if you are building an ap
     <div class="comment-date">2015-10-08 12:06 UTC</div>
   </div>
 </div>
-<div id="comments">
-    <hr>
-    <h2 id="comments-header">
-        Comments
-    </h2>
-    <div class="comment" id="c0b2e0bd555b5d5c5555c60bf11bff69">
-        <div class="comment-author"><a href="https://github.com/AmiradelBeyg">AmirB</a> <a href="#0AFE67B375254FE193A3FD10234A1CE9">#</a></div>
-        <div class="comment-content">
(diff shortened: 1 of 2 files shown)


ploeh:
Thank you for writing. The comment system here has its quirks, so I'm [tolerant in what I expect](https://en.wikipedia.org/wiki/Robustness_principle), so there's no need to worry if you've done everything perfectly. I'll fix any minor issues if there are any.

I wonder, however, if you meant to comment on [Service Locator is an Anti-Pattern](https://blog.ploeh.dk/2010/02/03/ServiceLocatorisanAnti-Pattern), which is what this pull request does.

Did you mean to instead comment on [Replacing Mock and Stub with a Fake](https://blog.ploeh.dk/2023/08/14/replacing-mock-and-stub-with-a-fake)?

While I'm guessing that this is the case, I don't wish to assume anything. I can easily make the edit on my side, but I didn't want to do it before asking if that's what you had in mind.

AmiradelBeyg:
Thank you Mark for your quick response. It looks like I might have accidentally tinkered with the wrong file—it was probably one of those late-night coding adventures. 😄

Anyway, I've corrected the comment now, and fingers crossed that everything else is A-OK! Thanks again for your patient.

ploeh:
Thank you. It's a great question, but it may take a few weeks before I can get around to posting an answer. I'll leave a little notification here once I do.

ploeh:
Thank you for writing. I've left [a short response on the page](https://blog.ploeh.dk/2023/08/14/replacing-mock-and-stub-with-a-fake/#d0ac4c5ecf444fb2a18b993ca16822de).

PR #945, opened 2024-06-05
Update 2024-06-03-youll-regret-using-natural-keys.html

nicholas-peterson:
Added comment about some plus sides to natural keys.


ploeh:
Thank you for writing. I've left [a response on the page](https://blog.ploeh.dk/2024/06/03/youll-regret-using-natural-keys/#2960b65bbaec4db8ade70f551e3f5062).

PR #948, opened 2024-06-07
Two keys are good

bantling:
Adding my thoughts on having two id keys in tables


ploeh:
Thank you for writing. I've left [a response on the page](https://blog.ploeh.dk/2024/06/03/youll-regret-using-natural-keys/#3ef814a896f34b5485aeeea739766fa9).

PR #952, opened 2024-06-15
Add reader comment

Marken-Foo:
Add comment on 2024-06-12-simpler-encapsulation-with-immutability


ploeh:
Thank you for writing. I'm quite hung up for next couple of weeks, so it may take some time before I get around to writing a response.

ploeh:
Thank you for writing. I've left [a response on the page](https://blog.ploeh.dk/2024/06/12/simpler-encapsulation-with-immutability/#7f2ed4d386144b378cae2206e8269a6d).

PR #957, opened 2024-07-22
Add comment for natural keys

jfheins:
Hi,
I thought I may add one of my experiences here. While a join table might not be the focus of the post, you never know when a join table might start to become more relevant.


ploeh:
Thank you for writing. I've left [a response on the page](https://blog.ploeh.dk/2024/06/03/youll-regret-using-natural-keys/#c573d77b80904d64b697418a7d4440cd).

PR #963, opened 2024-08-19
Add article commit

JesHansen:
Add a comment with question to article published 2024-08-12


ploeh:
Thank you for writing. I've left [a response on the page](https://blog.ploeh.dk/2024/08/12/using-only-a-domain-model-to-persist-restaurant-table-configurations/#3b2bacb8fa274590ad45cb7864f19848).

PR #965, opened 2024-10-15
Comment on Das verflixte Hunde-Spiel

anka-213:
I can't say that this is the most convenient way to write comments, but it definitely has its advantages. It also made me learn about the VSCode [Remote repositories extension](https://marketplace.visualstudio.com/items?itemName=GitHub.remotehub&WT.mc_id=devcloud-18509-cxa), so I wouldn't have to clone your repo just to write a comment.


ploeh:
Thank you for writing. I've left [a response on the page](https://blog.ploeh.dk/2024/10/03/das-verflixte-hunde-spiel/#984fc5acb2314c79b2f2d7ddfacea285).

anka-213:
yes, the ordering does indeed not make a large difference on its own (theoretically at most a factor of two with my algorithm). the primary improvement that changes the asymptomatics is to delay choosing the ordering of tiles as much as possible. for example, the first two tiles will be the same among (n-2)! permutations, so if we start by checking if the choice of first two tiles is valid, we don’t need to check any of those permutations at all, and so on recursively after each tile added.

so instead of trying all the possible 9! = 362,880 permutations and then rotations on top of that, 
we only explore 9,332 boards total, including rotations. I got the number experimentally by modifying the code to count how many boards it explored.

Issue #970, opened 2024-12-01
Don't index page 1, page 2, ..., page n


ploeh:
As a standard blog site, _ploeh blog_ has a set of [pages](https://blog.ploeh.dk/pages/) that a user may navigate using _next_ and _previous_ buttons. I don't think that  the average user would use that feature much, but according to Google Analytics, [Pages](https://blog.ploeh.dk/pages/) is ranked number 31 on the site.

The top page on the site (ranked 1) is, perhaps not surprising, the 'home page' at https://blog.ploeh.dk.

All that said, I sometimes need to find stuff on the site, and while I often search the source code (i.e. the HTML files), I also occasionally use a site-specific web search, and I've noticed that web search results often list, say [page 18](https://blog.ploeh.dk/page18) or [page 58](https://blog.ploeh.dk/page58/), simply because the crawler found a particular keyword on that page at that time.

These pages are 'aggregation pages', and articles move around on these pages as they get pushed further into the past. Therefore these search results aren't useful.

What's the best way to tell search engines to not index these pages? robots.txt?

I'm not up to date with modern SEO techniques, so would appreciate input if `robots.txt` isn't the best option.

arthurgubaidullin:
Hi @ploeh,

I don't think it's a good idea to disable indexing in your case, as there's no sitemap available for the search engine. Instead, I suggest adding canonical URLs to the meta information of the relevant pages. That should ideally be enough to address the issue.

Here's a helpful [link](https://developers.google.com/search/docs/crawling-indexing/consolidate-duplicate-urls#rel-canonical-link-method).

Best of luck with your article writing!

michalfi:
Maybe a wild idea, but a different angle of approach: As your blog has pretty sustained rate of new content, why don't you change paging to be time-invariant?

E.g. having a page per month. Your homepage could always show posts from the current month plus the last one (to ensure there's always at least a month of content), then a link to "October 2024", etc.

Or, maybe even simpler, just count the pages from the end?

ploeh:
> I don't think it's a good idea to disable indexing in your case, as there's no sitemap available for the search engine.

Would adding one help? After all, the [archive](https://blog.ploeh.dk/archive) contains almost everything of interest on the site, apart from the [About page](https://blog.ploeh.dk/about) and perhaps a few other pages.

[...]

PR #971, opened 2024-12-05
Adds canonical link to all post pages

arthurgubaidullin:
Adds canonical URLs to all post pages.

Relate to: #970


ploeh:
Thank you for [the explanation](https://github.com/ploeh/ploeh.github.com/issues/970#issuecomment-2521304943) over at #970. I can't say that I fully understand what this entails, but this pull request looks benign, so we'll see what happens.

(Sorry if this sounds a bit paranoid, but I'm generally being quite cautious with this site, since it now represents almost two decades of work.)

PR #972, opened 2025-01-01
Add comment to Pytest is Fast article

danielt1263:


ploeh:
Thank you for writing. I've left [a response on the page](https://blog.ploeh.dk/2024/12/30/pytest-is-fast/#d9b64e35daa34be7b5a5c34c55043583).

PR #975, opened 2025-03-10
Fix eggcorn reign->rein

bartelink:
Common usage will probably flip it to reigns over time but for now we get to imagine horses!


ploeh:
I should have caught that myself, but no-one is perfect 😀 Thank you 👍

PR #997, opened 2026-08-13
added comment

qfilip:
My vscode formatter took over, so the entire page is probably changed. Sorry about that.

### DIFF
--- a/_posts/2026-08-13-why-use-orms-if-llms-write-code.html
+++ b/_posts/2026-08-13-why-use-orms-if-llms-write-code.html
@@ -5,40 +5,158 @@ description: "The return of SQL?"
 date: 2026-08-13 6:47 UTC
 tags: [AI]
 ---
+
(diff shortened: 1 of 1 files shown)


ploeh:
> My vscode formatter took over, so the entire page is probably changed. Sorry about that.

is this some new thing? I just had to deal with the same issue in 664fbdb2, and in the 13 years I've been using my weird commenting system, I can't recall that it's happened before.

Since you've evidently taken time to write a good and thoughtful comment, I've fixed the issue myself, but I would appreciate if people wouldn't touch the main text in the future.

When I have time, I'll write a proper response to your comment, but this may take some days. I'll write back when I have an answer.

ploeh:
Sorry about the tardy response, but I've now added [a response to the page](https://blog.ploeh.dk/2026/08/13/why-use-orms-if-llms-write-code/#4d84baf78fef41c288b802df33a5895b).

PR #1001, opened 2026-08-22
Add comment to 2026-07-23-does-code-quality-still-matter

panesofglass:


ploeh:
Thank you for writing. I'll write back once I get around to adding a response.

A more immediate concern, however, that you may want to address: You've given your name as 'name', which I suppose is an oversight. If you want to change it, please send me a new pull request, or alternatively just reply here, and I can edit it on your behalf.

panesofglass:
Thank you for catching that! My name is Ryan Riley. I'll try to get to a PR later today unless you get to it first. Thank you!

